PR #9, opened 2021-08-29
Add option to ignore lines after headings

gavinmn:
This PR adds a new rule, `Heading blank lines top` that adds a blank line above every `# Heading`.

It's a very small modification of the `Heading blank lines` rule. I wasn't able to get it working as an option under that rule, but I'm open to suggestions as to how to do that if you want!


platers:
Thanks for the PR! I would prefer it as an option of `Heading blank lines`, maybe `top=true`. I think you can do it by just putting the `// trim blank lines before headings` line in an if. Look at YAML timestamp for an example of options. 

Also can you add the yarn.lock file to the .gitignore?

gavinmn:
Alright I gave making it an option under the `Heading blank lines` rule. Setting `bottom="false"` should now ignore adding spaces below headings. 

I wasn't entirely sure how to test the option with `npm test`, but the logic stayed the same. Let me know if you need any more changes or feel free to make any adjustments yourself!

Issue #36, opened 2021-09-12
Rule Suggestion: Convert between Tabs and spaces


chrisgrieser:
First of all, thanks for the plugin! It's one of those little things saving a bit of time every day. 

A rule that I would find very useful is the conversion between tabs and spaces. For me personally, this would be incredibly useful because all the notes I imported from earlier apps use spaces, but the various relationship line css snippets need tabs to work.

platers:
Would this logic work?

For every line, count the number of leading spaces, `spaces`. Replace them with `spaces/tabsize` tabs.

chrisgrieser:
I think so. 

One consideration could be to also deal with cases where you have one more space than your tabsize, which would result in `tab+space`. But that would rather be another rule, I guess?

chrisgrieser:
> You can pass it in as an option, the user can set `tabsize=2` in the settings. See the other rules for examples.

That would work, but it isn't ideal. Users might not consider that the tabsize set in the Linter plugin is independent from the tabsize set in the Obsidian main settings.

platers:
I was thinking about your use case, when you are copying from external sources. It would be nice if the default tabsize was the obsidian tab size though.

platers:
on the discord, found vault.getConfig("tabSize").

chrisgrieser:
is this how you would do it in the `rules.ts`?
```typescript
(text: string) => {
        let tabsize = vault.getConfig("tabSize"); // <- is this I would have to do it?
        (...)
},
```

also, concerning the examples, how exactly do I have to write them, with the indention in the code code of the `rules.ts` itself?

platers:
You should test the tabSize syntax in obsidian. If you put the repo in your plugins folder, after you run `npm run compile` you should be able to activate it in obsidian and test it.

Yep dedent deals with it, its pretty smart about the indentation.

chrisgrieser:
okay, I have a working solution ready, but there are two problem.

---

**Problem 1:** `vault.getConfig("tabSize")` does not work, when compiling, I get the following error:
```bash
(!) Plugin typescript: @rollup/plugin-typescript TS2304: Cannot find name 'vault'.
src/rules.ts: (654:23)

654         let tabsize = vault.getConfig("tabSize");
                          ~~~~~
```
I also tried adding `import { App } from 'obsidian';` to the imports at the top of the document, but that did not solve it. 

---

[...]

PR #52, opened 2021-09-18
2 New Rules + 2 Bugfixes

chrisgrieser:
- new rule to convert spaces to tabs, for now determining tabsize via option (see #36)
- new rule to ensure there is a line break at the end of a document
- fixed `format-tags-in-yaml` to also work when there are `,` between the original tags. (see added second example)
- fixed title case rule keeps the casing of the ignore-list-words instead of lowercasing them.


platers:
Can you merge? I changed how options work, it shouldn't be a difficult change.

chrisgrieser:
tbh, I have never resolved conflict in PRs before and looking through the code, it seems some stuff was moved to totally wrong places by the GitHub auto-merging function (?). I am closing this PR and inserting my stuff into a new PR, seems to be easier :/

PR #57, opened 2021-09-19
2 new rules, 2 bugfixes (re-submit)

chrisgrieser:
okay, next try. Like the last time:
- new rule for converting spaces to tabs (#36)
- new rule for ensuring there is a line break at the end of the document
- bugfix for titlecasing
- bugfix for YAML header tags with commata

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -1,7 +1,7 @@
 import dedent from 'ts-dedent';
 import moment from 'moment';
 import {headerRegex, ignoreCodeBlocksAndYAML, initYAML, insert} from './utils';
-import {Option, BooleanOption, MomentFormatOption} from './option';
+import {Option, BooleanOption, MomentFormatOption, TextOption} from './option';
 
 export type Options = { [optionName: string]: any };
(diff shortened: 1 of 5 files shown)


src/rules.ts
```diff
@@ -717,6 +733,64 @@ export const rules: Rule[] = [
...
+            	- text indented with 3 spaces
+            - text with no indention
+            		- text indented with 6 spaces
+        `,
+            /* eslint-enable no-mixed-spaces-and-tabs, no-tabs */
+            {tabsize: '3'},
+        ),
+      ],
+      [
+        new Option('Tabsize', 'Number of Spaces that will be converted to a Tab (default: 4)', '4'),
```
platers, line 770: I forgot to implement IntegerOptions, do you mind if I push to your branch? I can also push it to master, and then you will have to merge again.

chrisgrieser: no problem, you can simply push it to my branch, seems like less work

chrisgrieser: or you could accept my PR into the master and then simply InterOptions to the master? Wouldn't that be even easier?

PR #72, opened 2021-09-24
fixes `move-footnotes-to-the-bottom`

### LINKED ISSUE #71: A Footnote Text is Added
This time, a footnote text was added after I had executed the Linter plugin.

Here is the initial note:
image

Here is the modified one:
image

See the uploaded files in https://discord.com/channels/686053708261228577/722584061087842365/890900413547696139

chrisgrieser:
should now properly fix #71

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -116,7 +116,7 @@ export const rules: Rule[] = [
       'Removes extra spaces after every line.',
       RuleType.SPACING,
       (text: string, options = {}) => {
-        if (options['TwoSpaceLinebreak'] === false) {
+        if (options['Two Space Linebreak'] === false) {
           return text.replace(/[ \t]+$/gm, '');
         } else {
           text = text.replace(/\b[ \t]$/gm, ''); // one whitespace
@@ -136,18 +136,18 @@ export const rules: Rule[] = [
         line with trailing spaces and tabs`,
         ),
         new Example(
-            'With `TwoSpaceLinebreak=true`',
+            'With `Two Space Linebreak = true`',
             dedent`
         # H1
         line with trailing spaces and tabs  `,
             dedent`
         # H1
         line with trailing spaces and tabs  `,
-            {TwoSpaceLinebreak: true},
+            {'Two Space Linebreak': true},
         ),
       ],
       [
-        new BooleanOption('TwoSpaceLinebreak', 'Ignore two spaces followed by a line break ("Two Space Rule").', false),
+        new BooleanOption('Two Space Linebreak', 'Ignore two spaces followed by a line break ("Two Space Rule").', false),
       ],
   ),
   new Rule(
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -115,19 +115,39 @@ export const rules: Rule[] = [
       'Trailing spaces',
       'Removes extra spaces after every line.',
       RuleType.SPACING,
-      (text: string) => {
-        return text.replace(/[ \t]+$/gm, '');
+      (text: string, options = {}) => {
+        if (options['TwoSpaceLinebreak'] === false) {
+          return text.replace(/[ \t]+$/gm, '');
+        } else {
+          text = text.replace(/\b[ \t]$/gm, ''); // one whitespace
+          text = text.replace(/\b[ \t]{3,}$/gm, ''); // three or more whitespaces
+          text = text.replace(/\b( ?\t ?)$/gm, ''); // two whitespaces with at least one tab
+          return text;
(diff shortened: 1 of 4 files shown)


chrisgrieser:
- should fix #73 , if it isn't based on some weird rule interaction
- added new rule: `footnote-after-punctuation`
- added option to `trailing-spaces` to keep [Two-Space-Linebreaks](https://stackoverflow.com/questions/35177253/why-markdown-syntax-requires-two-spaces-before-n-to-make-a-paragraph-out-of-it#:~:text=1%20Answer&text=It%20doesn't%20require%20two,%2C%20i.e.%20tags.).

src/rules.ts
```diff
@@ -115,19 +115,39 @@ export const rules: Rule[] = [
...
+        # H1
+        line with trailing spaces and tabs  `,
+            dedent`
+        # H1
+        line with trailing spaces and tabs  `,
+            {TwoSpaceLinebreak: true},
+        ),
+      ],
+      [
+        new BooleanOption('TwoSpaceLinebreak', 'Ignore two spaces followed by a line break ("Two Space Rule").', false),
```
platers, line 150: can rename to `Two Space Linebreak` for readability

src/rules.ts
```diff
@@ -115,19 +115,39 @@ export const rules: Rule[] = [
...
       RuleType.SPACING,
-      (text: string) => {
-        return text.replace(/[ \t]+$/gm, '');
+      (text: string, options = {}) => {
+        if (options['TwoSpaceLinebreak'] === false) {
+          return text.replace(/[ \t]+$/gm, '');
+        } else {
+          text = text.replace(/\b[ \t]$/gm, ''); // one whitespace
+          text = text.replace(/\b[ \t]{3,}$/gm, ''); // three or more whitespaces
+          text = text.replace(/\b( ?\t ?)$/gm, ''); // two whitespaces with at least one tab
```
platers, line 124: can you add tests for each of these cases in test.ts? Im not the best at verifying regex's.

src/rules.ts
```diff
@@ -698,15 +718,17 @@ export const rules: Rule[] = [
...
 
             // remove footnotes directly before/after a line of text
-            text = text.replace(/\n?\n\[\^\w+\]: .*\n\n?/gm, '\n');
+            text = text.replace(/\n?(?:\n\[\^\w+\]: .*)+\n?/gm, '\n');
 
-            // remove footnotes sourrounded by text
-            text = text.replace(/\n\[\^\w+\]: .*\n/gm, '');
-            text += '\n\n' + footnotes.join('\n'); // append footnotes at the very end of the note
+            // append at the bottom
+            text = text.replace(/\n*$/, '') + '\n\n' + footnotes.join('\n'); // append footnotes at the very end of the note
```
platers, line 731: I assume these changes are to fix a bug? Can you add the previously failing test to tests.ts?

chrisgrieser: yeah, they are bug fixes for #71.

src/rules.ts
chrisgrieser: will do

src/rules.ts
chrisgrieser: to be honest, I have no idea what exactly those tests do other than somehow checking that the before-example matches the after-example when the rule is applied. So I do not really know what I would have to do when you say "add tests for each of these cases" :(

(Sorry, as I once mentioned, I only know some basic coding I thought myself)

src/rules.ts
chrisgrieser: same as above, I lack knowledge about tests to do that, I am afraid

src/rules.ts
platers: No problem :)
The tests are exactly as you said, the check that applying a rule to the before text makes it into the after text. The format is very similar to an example, except they dont show up in the docs to avoid cluttering them. The purpose of the tests is to make sure edge cases are handled correctly.

So in this case, since there are no tests for this rule yet, add a describe statement to put tests into at https://github.com/platers/obsidian-linter/blob/master/src/test.ts#L256. 
Then add `it` statements to specify a behavior that the rule should have. Here you could add an it statement `it('Ignores 2 trailing spaces'...` with the before text containing lines with one whitespace, 3 or more whitespaces, and 2 whitespaces. This way we can verify the regex's work as intended. (basically its just a hidden example)

Let me know if you have any questions

chrisgrieser:
Alright, it took me a bit, but I think I figured it out and was able to add a bunch of tests, which the updated rules all pass

PR #85, opened 2021-09-28
New Rule: Remove Multiple Spaces

chrisgrieser:
- only effects spaces between words
- spaces next to `-` (ul) are ignored (are covered by `space-after-list-markers`)
- spaces at the end are ignored (covered by `trailing-spaces`)
- spaces at the beginning (indention) are ignored as well
- see https://regex101.com/r/6heqdq/1

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -400,7 +400,7 @@ export const rules: Rule[] = [
   ),
   new Rule(
       'Remove Multiple Spaces',
-      'Removes or multiple consecutive spaces. Ignores Spaces at the beginning and ending of the line. ',
+      'Removes two or more consecutive spaces. Ignores spaces at the beginning and ending of the line. ',
       RuleType.SPACING,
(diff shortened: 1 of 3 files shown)


src/rules.ts
```diff
@@ -397,6 +397,27 @@ export const rules: Rule[] = [
         ),
       ],
   ),
+  new Rule(
+      'Remove Multiple Spaces',
+      'Removes or multiple consecutive spaces. Ignores Spaces at the beginning and ending of the line. ',
```
platers, line 402: `Removes or multiple consecutive spaces.` - think the grammar is off.
spaces should be lowercase

platers (commented):
Thanks for the regex101 link, didn't know about that site!

src/rules.ts
chrisgrieser: fixed

Issue #102, opened 2021-10-12
[Feature] Configurable `date created` and `date modified` keys


JeppeKlitgaard:
I think it would be really neat if the keys for the `date created` and `date modified` attributes in the `YAML-timestamp` rule were configurable.

I prefer to have no spaces in the keys and have used `created` and `modified` as my keys so far, but this isn't compatible with `obsidian-linter`. A set of text inputs to define the keys would be a fantastic improvement in my opinion!

PR #103, opened 2021-10-12
Configurable timestamp keys and multiple save fix

### LINKED ISSUE #100: Minor bug: duplicate linting when linting on save
When using the "Lint on Save" setting and then saving a file, I get two notifications from the Linter plugin. As the notifications show different information, it seems that this is not a notification error but rather the linter running twice?

Obviously, running the Linter twice in a row isn't a big deal since the second run will simply do nothing, but at least the double notification is a bit misleading.

Screenshot 2021-10-11 15 19 28

and when saving, the console indeed reports two linter runs:

```
running linter main.ts:132
running linter main.ts:122
```
The problem does not occur when I use "lint the current file" instead of the save file command.

JeppeKlitgaard:
Fixes #100 and #102.

Has a small amount of chore/linting touchups too.

See commit messages for more info.

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -133,7 +133,7 @@ export const rules: Rule[] = [
         new Example(
             'Removes trailing spaces and tabs',
             dedent`
-        # H1   
+        # H1
         line with trailing spaces and tabs	        `, // eslint-disable-line no-tabs
             dedent`
         # H1
@@ -540,17 +540,24 @@ export const rules: Rule[] = [
         text = initYAML(text);
 
         return formatYAML(text, (text) => {
-          if (options['Date Created'] === true && !text.match(/\ndate created:.*\n/)) {
+          const created_match_str = `\n${options['Date Created Key']}.*\n`;
+          const created_match = new RegExp(created_match_str);
+
+          if (options['Date Created'] === true && !text.match(created_match)) {
             const yaml_end = text.indexOf('\n---');
             const formatted_date = moment(options['metadata: file created time']).format(options['Format']);
-            text = insert(text, yaml_end, `\ndate created: ${formatted_date}`);
+            text = insert(text, yaml_end, `\n${options['Date Created Key']}: ${formatted_date}`);
           }
+
+          const modified_match_str = `\n${options['Date Modified Key']}.*\n`;
+          const modified_match = new RegExp(modified_match_str);
+
           if (options['Date Modified'] === true) {
-            text = text.replace(/\ndate modified:.*\n/, '\n');
+            text = text.replace(modified_match, '\n');
             text = text.replace(/\ndate updated:.*\n/, '\n'); // for backwards compatibility
             const yaml_end = text.indexOf('\n---');
             const formatted_date = moment(options['metadata: file modified time']).format(options['Format']);
-            text = insert(text, yaml_end, `\ndate modified: ${formatted_date}`);
+            text = insert(text, yaml_end, `\n${options['Date Modified Key']}: ${formatted_date}`);
           }
           return text;
         });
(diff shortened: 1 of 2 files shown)


src/rules.ts
```diff
@@ -133,7 +133,7 @@ export const rules: Rule[] = [
         new Example(
             'Removes trailing spaces and tabs',
             dedent`
-        # H1
```
platers, line 136: revert the trailing spaces in this example

src/rules.ts
```diff
@@ -501,7 +501,7 @@ export const rules: Rule[] = [
 
           for (const line of insert_lines) {
             const key = line.split(':')[0];
-            if (!parsed_yaml.hasOwnProperty(key)) {
+            if (!Object.hasOwnProperty.call(parsed_yaml, key)) {
```
platers, line 504: I'm not a javascript expert, why do you use Object.hasOwnProperty?

src/rules.ts
```diff
@@ -593,7 +600,9 @@ export const rules: Rule[] = [
       ],
```
platers, line 600: can you add an example showing what the new options do?

src/test.ts
```diff
@@ -127,13 +127,13 @@ describe('Rules tests', () => {
       const before = dedent`
         Line
         - 1
-        -
```
platers, line 130: can you revert this line

src/test.ts
```diff
@@ -127,13 +127,13 @@ describe('Rules tests', () => {
...
         Line
         - 1
-        - 
+        -
         Line
         `;
       const after = dedent`
         Line
         - 1
-        -
```
platers, line 136: here too.

src/test.ts
```diff
@@ -317,26 +317,26 @@ describe('Trailing spaces', () => {
   it('Tab-Space-Linebreak removed', () => {
     const before = dedent`
         # H1
-        line with trailing tab and spaces
```
platers, line 320: all the tests and examples for the trailing spaces rule got messed up, can you revert those changes so they test removing trailing spaces?

src/test.ts
```diff
@@ -425,7 +425,20 @@ describe('yaml timestamp', () => {
     date created: 2019-01-01
     ---
     `;
-    expect(rulesDict['yaml-timestamp'].apply(before, {'Date Created': true})).toBe(after);
+    expect(rulesDict['yaml-timestamp'].apply(before, {'Date Created': true, 'Date Created Key': 'date created'})).toBe(after);
```
platers, line 428: this can be an example.
show the date modified key option just for completeness

platers (commented):
Thanks for the PR! Looks pretty good to me, some tests got messed up during linting.

src/rules.ts
JeppeKlitgaard: I am not super familiar in TS/JS either, but ESLint complained. I believe this is because it is considered 'bad' to assume an object inherits from the Object class and thus has a `hasOwnProperty` method.

There is a short explanation on the relevant ESLint rule: https://eslint.org/docs/rules/no-prototype-builtins

src/test.ts
JeppeKlitgaard: Ah, apologies. My editor automatically strips trailing whitespace and I didn't consider that the tests with trailing whitespace would obviously be affected. Good catch!

Fixed in new commit

src/rules.ts
JeppeKlitgaard: Fixed

src/rules.ts
JeppeKlitgaard: Example added

Issue #135, opened 2021-10-30
[Bug] Replaces $$ with $ in code blocks


JeppeKlitgaard:
Most of the rules seem to replace `$$` with `$` in fenced blocks.

This ruins the math markup.

PR #136, opened 2021-10-30
fix: dollar signs disappearing

JeppeKlitgaard:
See #135


platers:
LGTM, could you just add a test somewhere that previously failed?

JeppeKlitgaard:
There you go

PR #159, opened 2022-01-17
new rule `remove-empty-list-markers` & minor fix for `format-tags-in-yaml`

chrisgrieser:
`remove-empty-list-markers` will simply remove bullet points you have left empty.

---

The minor fix for `format-tags-in-yaml` is related to the fact, that yaml-keys with `-` were not matched with the current regex using `\w+`. This is most notable when using the breadcrumbs plugin, where you have a a yaml line like `BC-tag-note: '#coding'` where you explicitly *need* the `#`. By changing `\w+` to `[A-Za-z-]+`, I fixed that. You can see the difference here:
- with `\w+`: https://regex101.com/r/reyLZq/1
- with `[A-Za-z-]+`: https://regex101.com/r/0L4vd5/2


platers:
thanks!

Issue #165, opened 2022-01-25
Add demo gif to readme


platers:
Show a file being linted. Would help people understand what the plugin does. Would appreciate if someone recorded a short gif!

PR #169, opened 2022-02-12
add a short demo gif to README

SewerynKras:
closes https://github.com/platers/obsidian-linter/issues/165


platers:
Hey thanks for making this! If its not too much trouble, could you pause a bit longer on the original doc, its hard to tell how messy it is because you lint it so fast lol.

SewerynKras:
Sure, I made the before and after more less the same length. Do you think it looks good now?

platers:
thanks!

Issue #32, opened 2021-09-08
Add rules from markdownlint


platers:
https://github.com/DavidAnson/markdownlint

Add rules that are automatically fixable.

pjkaufman:
Looks like the following are the rules tat would be able to be fixable:

[MD004](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004) ul-style - Unordered list style
[MD005](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md005) list-indent - Inconsistent indentation for list items at the same level
[MD007](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md007) ul-indent - Unordered list indentation
[MD009](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md009) no-trailing-spaces - Trailing spaces
[MD010](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md010) no-hard-tabs - Hard tabs
[MD011](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md011) no-reversed-links - Reversed link syntax
[MD012](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md012) no-multiple-blanks - Multiple consecutive blank lines
[MD014](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md014) commands-show-output - Dollar signs used before commands without showing output
[MD018](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md018) no-missing-space-atx - No space after hash on atx style heading
[MD019](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md019) no-multiple-space-atx - Multiple spaces after hash on atx style heading
[MD020](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md020) no-missing-space-closed-atx - No space inside hashes on closed atx style heading
[MD021](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md021) no-multiple-space-closed-atx - Multiple spaces inside hashes on closed atx style heading
[MD022](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022) blanks-around-headings/blanks-around-headers - Headings should be surrounded by blank lines
[...]

PR #200, opened 2022-05-13
Ignore Links for Rules

### LINKED ISSUE #196: [Usability-breaking issue] Linter modifies links
I have files with multiple spaces in their names, because reasons. Using Linter breaks the links to all those files by removing the double spaces. I'm sure this applies to other filters, and is something that completely breaks the usability of the plugin.

And yes, I can remove the double spaces from the file names, but Linter absolutely shouldn't be messing with anything that's inside a `link` environment anyways.

pjkaufman:
@platers 

Fixes #196

I went ahead and added a the regex for links and the UTs to make sure the 3 kinds of links lint correctly. However a couple of situations did not work for me, but I am not sure if that was intended:
- `[  ]` does not have the extra spaces removed from it for some reason
- `(  )` also does not have the extra spaces removed from it for some reason

I added a links placeholder which had to have English letters so the English and Chinese spacing lint rule would still work.
I renamed `ignoreCodeBlocksAndYAML` to `ignoreCodeBlocksYAMLAndLinks` since it now does more things.

Please let me know if we need to not remove the empty spaces from the two above scenarios that I hit issues with or if there is a different way we would like to ignore the links before any rules are added around them (#32  is one that would require reworking this so that the rules for links would work just fine while the rules for other operations would not affect it).

I am not sure if remark has a link position option like for the code blocks, but if it does, please let me know and I can try to use that if we would like.


src/test.ts
```diff
@@ -674,3 +674,51 @@ describe('YAML Title', () => {
...
+    ![This has  spaces in it](File with  spaces.png)
+    `;
+
+    expect(rulesDict['trailing-spaces'].apply(before)).toBe(after);
+  });
+  it('Wiki link with spaces stays the same', () => {
+    const before = dedent`
+    # Hello world
+
+    [This has  spaces in it](File with  spaces.md)
```
platers, line 713: should this test `[[wiki    link]]`?

src/utils.ts
```diff
@@ -96,19 +96,23 @@ export function moveFootnotesToEnd(text: string) {
...
  * @return {string} The processed text
  */
-export function ignoreCodeBlocksAndYAML(text: string, func: (text: string) => string): string {
+export function ignoreCodeBlocksYAMLAndLinks(text: string, func: (text: string) => string): string {
   const codePlaceholder = 'PLACEHOLDER 321417';
   const ret = replaceCodeblocks(text, codePlaceholder);
   text = ret.text;
   const replacedCodeBlocks = ret.replacedCodeBlocks;
 
+  const linkPlaceHolder = 'PLACEHOLDER_LINK 57849';
```
platers, line 112: I think we should replace YAML before links, if there is a link in the YAML we dont want to touch it at all.

platers (commented):
thanks!

src/test.ts
pjkaufman: That is correct. I have updated the test.

src/utils.ts
pjkaufman: Sounds good. I have swapped the order.

platers (approved):

PR #218, opened 2022-06-04
Match Header Capitalization Style Title to Formatting Result

muya:
- Update the Heading Capitalization Style Options to match what they do
- Fix documentation for ALL CAPS dropdown option

### Before:
obsidian-linter-style-formatting-before

### After:
obsidian-linter-style-formatting-after


platers:
Thanks, nice implicit documentation.

PR #225, opened 2022-06-12
Timestamp YAML Rule Now Fills in Empty Keys and Only Updates Modified Key with a File Change

### LINKED ISSUE #221: Timestamp improvement
I have linting set to auto-lint on save.

As a habit, I always lint files, even if I did not modify. The issue is if there was no linting issue (the file was not changed), the modified date is updated regardless, even if nothing was actually modified.

It gets worse when I lint all files, because all files will have their modified timestamps changed. This is an issue esp when using version control, because this will modify all files all the time.

NB: I don't think anyone would want to change the timestamps of an unchanged file, so maybe it would be better to just add this improvement without a settings checkbox. Should be a minor version change.

pjkaufman:
Fixes #221 

There was an issue where if you were to lint a file that had no other changes besides the what would be added by date modified, it would update the file with date modified value. This should be fixed via this PR. 
I was also made aware that there was an issue where the keys for the timestamp YAML rule would not be filled in if they were present and blank (this became a new issue for updated date when it was only to be changed on an update to the file). This should be fixed now as well. We may want to consider cleaning up the logic here since it is getting a little clunky, but this should do for now.

Changes Made:
- Added tests to account for different update scenarios as well as blank values to help mitigate any issues with this in the future
- Updated the way we run rules to wait to run timestamp YAML until the very end so we can check if the file has changed more easily

@platers , could you elaborate on why we have logic for backwards compatibility around replacing the date modified and date created values? It would help a good bit if I could remove them since they seem to just be an extra replace something changed a while back (about 9 months ago): https://github.com/platers/obsidian-linter/blame/a423daddd57f4a0b40c99160a50a0b0028797997/src/rules.ts#L432

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -149,15 +149,16 @@ export default class LinterPlugin extends Plugin {
 
       // remove hashtags from tags before parsing yaml
       const tag_rule = rules.find((rule) => rule.name === 'Format Tags in YAML');
-      const options = tag_rule.getOptions(this.settings);
-      if (options[tag_rule.enabledOptionName()]) {
-        newText = tag_rule.apply(newText, options);
+      const tag_options = tag_rule.getOptions(this.settings);
+      if (tag_options[tag_rule.enabledOptionName()]) {
+        newText = tag_rule.apply(newText, tag_options);
       }
 
       const disabledRules = getDisabledRules(newText);
 
       for (const rule of rules) {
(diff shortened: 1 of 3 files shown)


platers:
I think we can remove that line now

pjkaufman:
> I think we can remove that line now

Sounds good. I'll remove it and then merge the PR.

Issue #236, opened 2022-06-24
Bug: remove-link-spacing removes spacing inside checkbox when followed by markdown link


olearydj:
Hello again.
I noticed today that when `remove-link-spacing` is active, empty check boxes (`- [ ]`) get nuked on lint. I don't use them often but when I do, I they get linted into oblivion. A sub-option on this to exclude checkboxes would be great.

pjkaufman:
@olearydj , what do you mean by get nuked? Could you provide an example as well as a copy of your `data.json` for this plugin to help me reproduce the issue?

olearydj:
Sorry, I just mean that the linter removes the space between the square brackets, which breaks the checkbox.

pjkaufman:
I can try a couple more rule combinations,but without the list of rukes you have active (the `data.json`) and/or a snippet of markdown with which to reproduce the situation with, I may not be able to recreate and fix the scenario you are describing.

I have not seen the behavior mentioned as of yet.

Here ias what I believe you are saying is happening:
``` markdown
- [ ] 

Other content here...
```
After lint:
``` markdown
- [] 

Other content here...
```

Is this the scenario you are describing, @olearydj ?

pjkaufman:
Just in case you are not aware of where the `data.json` is, it should be located in `path-to-your-vault/.obsidian/plugins/obsidian-linter/`.

pjkaufman:
I may have found a way to reproduce this issue, but I will have to do some more testing to see if I am actually able to do so. If I can, I will see about fixing this for this release.

pjkaufman:
So it looks like if you have a link in a checkbox, it can cause an issue as follows.
Before:
``` markdown
- [ ] [Songs Tasks](Tasks/SongsTasks.md)
```
After:
``` markdown
- [] [Songs Tasks](Tasks/SongsTasks.md)
```

This seems like it will not be too hard to fix. It is likely due to not excluding a particular character from the regex. I will see about fixing this once I finish going through the changes made by the linting all my personal files to see if any issues popped up.

olearydj:
Yes, that was exactly the situation. It was a list of obsidian plugins to look into, actually. Todo list of links.
Sorry I wasn't more specific / thorough in my reporting this time.

I've updated the plugin and can confirm this problem is fixed where I was able to repro it before.

Thank you!

PR #237, opened 2022-06-25
Escape YAML Where possible

### LINKED ISSUE #224: FR: wrap yaml strings in quotes as necessary
For example, I use a title field in my yaml headers. Books and academic papers commonly include colons. When LINTER encounters this it fails silently - nothing happens (no error notification) but the console shows YAMLException. Simply wrapping the offending text in quotes fixes this.

Matched and unmatched single (') or double (") quotes also cause problems with interpreting the yaml string. Ideally you could recognize those cases and wrap the whole value with the appropriate (opposite) quote mark.

I really love this plugin - it has saved me a TON of time!! Thank you for the work.

ps - my first issue / FR on GH so please excuse any shortcomings

pjkaufman:
Fixes #224 

This is meant to help reduce issues where the linter is not able to run because it needs to parse the YAML due to a YAML rule being active, yet the YAML is invalid and can be fixed with a simple escape by single or double quotes. Some scenarios cannot be remedied, but hopefully this will catch most use cases.

Do note that using the try single line arrays option will not work properly if you nest a comma inside quotes since it will be treated as the delimiter despite being escaped.

Changes Made:
- Removed some ghost code
- Added a new rule for attempting to escape YAML special characters
- Added tests to verify that the changes work as expected
- Updated the linter rule running logic to make sure if the rule is active it gets run first and that only rules that are not disabled, run prior, or run after the regular rules get run in the for loop (this will hopefully save some time when linting hundreds to thousands of files at once).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Documentation for all rules can be found in the [rules docs](https://github.com/
 - [insert-yaml-attributes](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#insert-yaml-attributes)
 - [yaml-timestamp](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#yaml-timestamp)
 - [yaml-title](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#yaml-title)
-- [escape-special-special-characters](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#escape-special-special-characters)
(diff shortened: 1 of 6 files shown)


pjkaufman:
The YAML changes run against several thousand of my one files looks good. I am not seeing any weird issues around it. It has however shown me that there may be an issue with a adding 2 spaces after content. I will take a closer look at it and see if I can fix it if it was not just me creating a bad YAML setup. It also seems that I may be able to reproduce #236 , but I will have to test it out a little more to be certain.

PR #256, opened 2022-07-04
Add rule YAML Title Alias

### LINKED ISSUE #254: Rule: YAML alias from title
The rule should transform

```
# My title
```

into

```
---
aliases:
  - My title
---

# My title
```

This would be helpful as `Quick Switcher` (`Ctrl + O`) supports aliases, so notes could be found not only by filenames but also with titles

***

However, I am not sure how to handle edits

```
---
aliases:
  - Old title
---

# Old title
```

edited to

```
---
aliases:
  - Old title
---

# New title
```

How can we possibly detect that `Old title` was the previous title on `modify` event?

mnaoumov:
Fixes #254

### DIFF
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -17,6 +17,8 @@ import {
   removeSpacesInLinkText,
   toSingleLineArrayYamlString,
   setYamlSection,
+  getYamlSectionValue,
+  removeYamlSection,
 } from './utils';
 import {
   Option,
@@ -1679,18 +1681,15 @@ export const rules: Rule[] = [
 
         let yaml = text.match(yamlRegex)[1];
 
-        const previousTitleMatch = yaml.match(/(?<=^|\n)linter-yaml-title-alias:[ \t]+(.+)/);
-
-        let previousTitle = previousTitleMatch == null ? null : loadYAML(previousTitleMatch[1]);
+        let previousTitle = loadYAML(getYamlSectionValue(yaml, 'linter-yaml-title-alias'));
 
         if ((previousTitle === title && !shouldRemoveTitleAlias) || (previousTitle === null && shouldRemoveTitleAlias)) {
           return text;
         }
 
-        const aliasesRegex = /(?<=^|\n)aliases:[ \t]*(\S.*|(?:\n {2}\S.*)+)\n(?=$|\S)/;
-        let aliasesMatch = yaml.match(aliasesRegex);
+        let aliasesValue = getYamlSectionValue(yaml, 'aliases');
 
-        if (!aliasesMatch) {
+        if (!aliasesValue) {
           if (shouldRemoveTitleAlias) {
             return text;
           }
@@ -1716,11 +1715,9 @@ export const rules: Rule[] = [
 
           text = text.replace(`---\n${yaml}---`, `---\n${newYaml}---`);
           yaml = newYaml;
-          aliasesMatch = yaml.match(aliasesRegex);
+          aliasesValue = getYamlSectionValue(yaml, 'aliases');
           previousTitle = '';
         }
-
-        const aliasesValue = aliasesMatch[1];
 
         const isMultiline = aliasesValue.includes('\n');
         const parsedAliases = loadYAML(aliasesValue);
@@ -1763,7 +1760,7 @@ export const rules: Rule[] = [
         newYaml = setYamlSection(newYaml, 'aliases', newAliasesYaml);
 
         if (shouldRemoveTitleAlias) {
-          newYaml = newYaml.replace(/(?<=^|\n)linter-yaml-title-alias:.+\n/, '');
+          newYaml = removeYamlSection(newYaml, 'linter-yaml-title-alias');
         } else {
           newYaml = setYamlSection(newYaml, 'linter-yaml-title-alias', ` ${toYamlString(title)}`);
         }
--- a/src/rules.ts
+++ b/src/rules.ts
@@ -1667,6 +1667,9 @@ export const rules: Rule[] = [
       'Inserts the title of the file into the YAML frontmatter\'s aliases section. Gets the title from the first H1 or filename.',
       RuleType.YAML,
       (text: string, options = {}) => {
+        const ALIASES_YAML_SECTION_NAME = 'aliases';
+        const LINTER_ALIASES_HELPER_NAME = 'linter-yaml-title-alias';
+
         text = initYAML(text);
         let title = ignoreCodeBlocksYAMLTagsAndLinks(text, (text) => {
           const result = text.match(/^#\s+(.*)/m);
@@ -1681,13 +1684,13 @@ export const rules: Rule[] = [
 
         let yaml = text.match(yamlRegex)[1];
 
-        let previousTitle = loadYAML(getYamlSectionValue(yaml, 'linter-yaml-title-alias'));
+        let previousTitle = loadYAML(getYamlSectionValue(yaml, LINTER_ALIASES_HELPER_NAME));
 
         if ((previousTitle === title && !shouldRemoveTitleAlias) || (previousTitle === null && shouldRemoveTitleAlias)) {
           return text;
         }
 
-        let aliasesValue = getYamlSectionValue(yaml, 'aliases');
+        let aliasesValue = getYamlSectionValue(yaml, ALIASES_YAML_SECTION_NAME);
 
         if (!aliasesValue) {
           if (shouldRemoveTitleAlias) {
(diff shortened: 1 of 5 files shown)


pjkaufman:
@mnaoumov , thabks for creating a PR for this. I just had a couple of questions to make sure I understand how this is supposed to work.

It looks like it forces aliase tags to be multiline arrays no matter what. Am I mistaken or does it actually force the change on the alias key?

How does it handle titles with a colon and a space, a single quote, or a double quote? Is the alias escaped to prevent bad YAML from being created?

Why change the `escapeYAMLString` method param to an object of type any? Was that because tou were reading in a value from the options?

Overall, it looks pretty good from my once over. I will take a more thorough look later.

pjkaufman:
I took a closer look at one of the examples and it looks invalid to me:

Example: Adds a header with the title.

Before:

```markdown

```

After:

```markdown
---
aliases:
  # linter-yaml-title-alias
  - Filename
---

Is there a reason a markdown header is in the frontmatter? This is invalid yaml from my understanding.

**Comment by mnaoumov** [comment]
@pjkaufman yes, it forces `aliases` to be a multiline array.

`# linter-yaml-title-alias` is not a header, it's a YAML comment. It's needed to mark which of the aliases has to be replaced in case if header changed. Without this comment we can't distinguish which alias represents old title that has to be replaced vs some other alias that has to be kept.

To denote this functionality, I have two different unit tests: `Updates changed alias` and `Adds title before existing aliases array`

Special characters such a colons are handled property and there is a corresponding unit test for that.

`escapeYAMLString` was changed because it turned out that it can do more than just escaping. Actually it's a YAML serializer. It converts any object into its YAML string representation. In my PR I needed to serialize some object back to the YAML string, that's why I used this function.

**Comment by pjkaufman** [comment]
@mnaoumov , is there a reason to not include the yaml comment after the list item like so:

``` yaml
---
aliases:
  - Filename # linter-yaml-title-alias
---
```
Also, are we able to move the forcing of the alias as a multiline array out into a separate rule since it may be undesired that the format of the array changes when using this rule?

mnaoumov:
@pjkaufman I rewrote my PR in order to preserve existing property styles. Corresponding tests are added

src/rules.ts
```diff
@@ -1658,6 +1659,168 @@ export const rules: Rule[] = [
...
+              emptyValue = ' [\'\']';
+              break;
+            case 'Single string':
+              emptyValue = ' \'\'';
+              break;
+            default:
+              throw new Error(`Unsupported setting 'YAML aliases new section style': ${options['YAML aliases new section style']}`);
+          }
+
+          const newAliasesSection = `aliases:${emptyValue} ${linterMarkerComment}\n`;
```
pjkaufman, line 1699: How does this work properly for `Single string` and `Single-line array`?

It looks like it will just add it at the end of the array which will need to be enforced and thus the comment is unnecessary. If it is the single line array it will likely add it at the end of the entry, but what keeps other values from being added after the comment? 

I guess all of that leads me to ask if it makes sense to not just enforce the order of where the alias will be in the list instead of using a comment since it seems to be ineffective for single-line arrays and single line strings?

src/rules.ts
```diff
@@ -1658,6 +1659,168 @@ export const rules: Rule[] = [
...
+          let newYaml = yaml.replace(/(?<=^|\n)aliases:\s*/, newAliasesSection);
+
+          if (newYaml === yaml) {
+            newYaml = `${newAliasesSection}${yaml}`;
+          }
```
pjkaufman, line 1704: Is this necessary since if the old yaml and new yaml are the same after replacing aliases then there is no need to make a change. Unless this is the case where  no `aliases` key is present. However the above looks to duplicate the `aliases`  key when nothing changes based on the rule.

src/rules.ts
```diff
@@ -1658,6 +1659,168 @@ export const rules: Rule[] = [
...
+        ),
+        new DropdownOption(
+            'YAML aliases new array style',
+            'The style of the aliases YAML property for newly created arrays',
+            'Multi-line array',
+            [
+              new DropdownRecord(
+                  'Multi-line array',
+                  '```aliases:\\n  - Title```',
+              ),
```
pjkaufman, line 1814: Why are there two options here? It looks like the first dropdown is sufficient.

src/rules.ts
pjkaufman: To expound on the above, if we are saying the first element in the array or the list of strings is the value is there not a better way to check if the value existed prior?

mnaoumov: This works properly, based on my unit tests. This piece of code is needed to create dummy aliases section if it doesn't exist or empty. It is being populated with proper values below. It is made like that to follow DRY principle

src/rules.ts
mnaoumov: The selected code is made to cover the case of completely missing `aliases` section.

The code above it it to cover the case of existing empty `aliases` section.

The corresponding unit tests are written for those cases

src/rules.ts
mnaoumov: No, first dropdown is used when `aliases` section is created from scratch.

The second dropdown is needed in the following scenario

1. First dropdown is set to be `Single string`
2. The note already has `Single string` `aliases` section
`
---
aliases: alias1
---
`
3. Now the linter has to write a new alias from title. So now we have two aliases, so we need to write an array, and we need to define is it going to be single-line or multi-line. Because of point 1, it's unclear, so we need another dropdown
`
---
aliases: [Title, alias1] # linter-yaml-title-alias
---
`

vs

`
---
aliases:
  - Title # linter-yaml-title-alias
  - alias1
---
`

src/rules.ts
mnaoumov: @pjkaufman I don't understand how are you going to check if the value existed prior? In the `modify` event of the note, we don't have an access to the previous version of the note, that's why I came up with this custom comment idea

pjkaufman: I am trying to wrap my head around it as well. I am not certain there is a better option. However it seems like it is just an indicator that the value existed prior.

pjkaufman: However, I did not see a test that checked what would happen if nothing had changed. That is a test case I am a little concerned about given the code I pointed out.

mnaoumov: Yes, you are totally right, as it just an indicator, I will rewrite it with the custom yaml property. It is way more reliable than the custom comment

src/rules.ts
pjkaufman: What if the aliases section does not need updating because it is already in the correct format?

src/rules.ts
pjkaufman: I don't think I follow why you cannot use the current style for the update. When you have a single string it is spaces which dictate what is an alias. For example,
``` yaml
aliases: alias1 alias2
```
should be the same as 
``` yaml
aliases: 
 - alias1 
 - alias2
```

src/rules.ts
mnaoumov: If it is already in the correct format, then this `if` check will just bypass the entire code block: https://github.com/platers/obsidian-linter/pull/256/files#diff-7704a77503c5e82d13beab70f6f18d85485e80fd5323703eecb4cce7b1edd3daR1685

src/rules.ts
pjkaufman: In order to keep an alias that has spaces in it as one, it should be inside of quotes. But maybe Obsidian does not work this way. I believe that it did, but I could be wrong.

mnaoumov: I am sorry, but this is not the way YAML works, `alias1 alias2` is a single string, it is the same as

```
aliases:
  - alias1 alias2
```

mnaoumov:
@pjkaufman I refactored my PR to use custom YAML key instead of the comment. It is way more powerful. By keeping the last key, we can quickly detect the no-op case. Also we no longer depend on the order of the aliases, which makes it less fragile

src/test.ts
```diff
@@ -1205,6 +1205,600 @@ describe('YAML Title', () => {
...
+    ---
+    aliases: alias1
+    linter-yaml-title-alias: alias1
+    ---
+    # Filename
+    `;
+
+    const after = dedent`
+    ---
+    aliases:
```
pjkaufman, line 1650: Should it remove the key as well rather than leaving it empty when dealing with a single string?

pjkaufman (commented):
Just a couple of questions before I approve. It looks pretty good. I would personally call the yaml section methods yaml key methods, but that is a minor thing. The only thing I am curious about are the two things I added review comments around.

src/rules.ts
pjkaufman: I tested it out and I must have seen things unless it was the tags section that allowed for weird functionality. I am aware that YAML does not allow it, but not all of what Obsidian does is valid YAML. That said, this is correct, I have no problem with this, but would like it if we could make it clear in the description of the drop down that the dropdown is meant for scenarios that have more than 1 alias.

src/rules.ts
pjkaufman: Sounds good.

pjkaufman:
Out of curiosity, what happens when you set the multivalue dropdown to single-line array, but your single value dropdown is set to single line? Does it go back to a single-line value when you only have one alias again?

For example
``` YAML
---
aliases:
  - alias 1
  - alias 2
---
```

has alias 2 removed and becomes:
``` YAML
---
aliases: alias 1
---
```

mnaoumov:
@pjkaufman I merged the dropdowns to clarify the intent

mnaoumov:
@pjkaufman I am considering to add a new boolean setting to keep existing style or force the configured one

mnaoumov:
> Just a couple of questions before I approve. It looks pretty good. I would personally call the yaml section methods yaml key methods, but that is a minor thing. The only thing I am curious about are the two things I added review comments around.

I called `yaml section` because key is just a string, but it can possibly be a large multi-line value, so I though `section` is the better term for that. Feel free to change the wording if you find it appropriate or consistent

pjkaufman (approved):
This looks good to me. Did you want to make any more changes before I merge it?

PR #263, opened 2022-07-08
Update Test Setup and Documentation

pjkaufman:
@platers , does breaking up the tests like this to make them more maintainable make sense to you?

The `test.ts` file was getting kind of large as we added more rules. To help address this, the tests were moved out to their own separate files in the `src/test` directory. There is a file for each rule that has extra tests, one for examples, one for empty values, and one for disabled rules. This should make it easier to know where tests are and how to add them without making a multi-thousand line file. I also updated the readme to include the new testing instructions.

Changes Made:
- Broke tests out into a file per rule tested, examples, disabled rules, and missing  rules
- Updated test based documentation
- Updated `caniuse-lite`


platers:
LGTM. 
Maybe we should move the developer section of the readme to a new file to avoid confusing users.

pjkaufman:
That sounds good to me. I will plan to do that soon. Hopefully later on today. That should help make it clearer what is for devs and what is for the average user.

Issue #262, opened 2022-07-07
Refactoring idea: Switch from strings to strongly-typed TypeScript constructs


mnaoumov:
While I was browsing the codebase I noticed a lot of places where options keys and its values are accessed by string, so typescript compiler can't detect typos if any.

I have some ideas how to improve it. But I know that authors rarely approve massive refactoring PRs. So I would like to get author's vision and approval before I start working on my PR.

pjkaufman:
@mnaoumov , do you mind showing a small example of a change you are thinking of? I would be open to such a change if it is intuitive and does not break the old setup (i.e. not able to be made backwards compatible with current `data.json`).

Right now I am really the only active maintainer. My background is not so much in JS, but I can make things work in it. I am definitely open to suggestions.

pjkaufman:
We could definitely do that for certain things that are common in the options object.

PR #264, opened 2022-07-08
[WIP] Refactoring to strongly-typed

mnaoumov:
PoC for #262


mnaoumov:
@pjkaufman as an example of my idea, I extracted one rule into its own class, and all usages of it and its settings are strongly typed. Moreover, this [commit](https://github.com/platers/obsidian-linter/pull/264/commits/be0edad2cd67b5519bcaaf0fa45fde353259bc7f) shows that the codebase had incorrect setting which was not covered by any tests

Also the tricks like [this](https://github.com/platers/obsidian-linter/pull/264/commits/e07bb2800caa5884289b2bd2de1af74bce0545e5) won't be longer needed as with my refactoring, the default values of the settings will be also set during unit tests and we won't need to have artificial checks for real `false` and just unset `undefined`

src/Rules/RuleBase.ts
```diff
@@ -0,0 +1,51 @@
...
+      throw new Error('get name() is not implemented');
+    }
+
+    get description(): string {
+      throw new Error('get description() is not implemented');
+    }
+
+    get type(): RuleType {
+      throw new Error('get type() is not implemented');
+    }
```
pjkaufman, line 25: Is there a reason to make these methods as opposed to letting them be properties? I guess you would have the benefit of getting an error when trying to access info when a `name`, `description` or `type` was not specified.

mnaoumov:
@platers in my opinion, adding all rules in one large file is kinda a violation of SRP and OCP. Having a separate self-contained class could even simplify the learning curve for the new contributors.

As I was saying in #262 I am concerned about making major refactorings without preliminary review and approval.

I am also thinking about extracting tests for one rule into a separate file as well

src/test.ts
```diff
@@ -504,55 +520,65 @@ describe('Convert spaces to tabs', () => {
...
-        line with three trailing spaces   
-      `;
-    const after = dedent`
-        # H1
-        line with three trailing spaces
-      `;
-    expect(rulesDict['trailing-spaces'].apply(before)).toBe(after);
+  ruleTest({
+    testName: 'One trailing space removed',
+    ruleClass: TrailingSpaces,
```
pjkaufman, line 525: I think I am missing how `TrailingSpaces` is converted over to an object rather than a type. Do you mind explaining how this works with the above logic that is as follows:
``` Typescript
function ruleTest<TOptions extends Options>(args: {
    testName: string,
    ruleClass: typeof RuleBase,
    before: string,
    after: string,
     options?: TOptions
  }) {
  it(args.testName, () => {
    const rule = args.ruleClass.rule;
    expect(rule.apply(args.before, args.options)).toBe(args.after);
  });
}
```
How does specifically the following work `args.ruleClass.rule;`? It is my understanding that an instance of the class needs to be made to access this property.

pjkaufman (commented):
The majority of the changes look great!

I am curios about a couple of things. Why separate `rules.ts` and `rules-list.ts`?
Is there a reason to change the naming conventions for the `Rules` folder? I would much prefer that we continued to use dashes between words in file names and that we do not uppercase letters in file names.

Beyond that, I think it looks pretty good except for a couple of areas I noted where my lack of understanding of JS/Typescript as a whole left me wondering how the logic worked.

It is my opinion at this time, that this would be an improvement in how the rules are setup and would help reduce development errors.

pjkaufman:
> @platers in my opinion, adding all rules in one large file is kinda a violation of SRP and OCP. Having a separate self-contained class could even simplify the learning curve for the new contributors.
> 
> As I was saying in #262 I am concerned about making major refactorings without preliminary review and approval.
> 
> I am also thinking about extracting tests for one rule into a separate file as well

@mnaoumov , I have already pulled out the tests into multiples files on the master branch (I just merged that change).

pjkaufman:
> Thanks for the POC! I'll take a closer look at this soon. My concern with adding abstraction is that it can make it harder for a new contributor to add a rule. Type safety is great though!

I share some of the same concern, but I do think that helping keep devs from accidentally referencing option names is a little bit of a larger concern for me personally. I like to have things separated out by files and have many smaller files rather than several larger files. Documentation in a markdown file for contributions could help new contributors better adjust to the way rules are created. I look forward to seeing what you think of this change when you take a closer look. I know for me some of it goes over my head, but most of it makes sense to me.

mnaoumov:
> Why separate `rules.ts` and `rules-list.ts`?

To avoid circular references. `rules.ts` has `Rule` class, which is used by `TrailingSpaces.ts`, but `TrailingSpaces.ts` is being used by `rules-list.ts`. If we don't separate them, we will have a circular reference.

> Is there a reason to change the naming conventions for the `Rules` folder?

No

>  I would much prefer that we continued to use dashes between words in file names and that we do not uppercase letters in file names.

I was mistakenly thinking that filename should match the class name. But I just checked, there is no such convention in TypeScript

> It is my opinion at this time, that this would be an improvement in how the rules are setup and would help reduce development errors.

In my opinion, your current string-based options is a very error-prone approach, which I proved in the commit mentioned above

src/Rules/RuleBase.ts
mnaoumov: `get name()` is not a method, it is a property getter. So the usage gonna be `const name = new TrailingSpaces().name`

Those properties are defined in the `RuleBase.ts` which is the base class and it kinda forces those properties to be overridden in the child classes. TypeScript doesn't have a concept of abstract methods such as in programming languages like Java or C#. So with this `throw new Error` approach we just give a runtime hint.

src/test.ts
pjkaufman: Gotcha. I was not aware that that was how we created Singletons in Typescript. That makes more sense now.

src/Rules/RuleBase.ts
pjkaufman: Thanks for this run down. I am gradually learning the nuances of Typescript, so this is helpful for me to learn.

mnaoumov:
@platers @pjkaufman I put my refactoring on pause. I don't want to invest my time if it has a low chance to be approved and merged. While those changes are going to be very repetitive, it is gonna be insanely difficult for you to do a code review for all of them. That's a dilemma

pjkaufman:
> @platers @pjkaufman I put my refactoring on pause. I don't want to invest my time if it has a low chance to be approved and merged. While those changes are going to be very repetitive, it is gonna be insanely difficult for you to do a code review for all of them. That's a dilemma

That makes sense. Once @platers  has a chance to take a closer look, we will have a better understanding of if we plan to go this direction. I am hopeful we will come to a solution that allows for maintainability and cuts down on errors in development.

pjkaufman:
@mnaoumov , FYI, looks like abstract methods were added in Typescript 1.6 according to the following : https://stackoverflow.com/a/13335645

mnaoumov:
@pjkaufman thanks a lot. I missed that part

platers:
It looks good to me though I'm also not familiar with some of the TS. Happy to move forward with this. Lets try to split the refactor into chunks when we can, but I'm not to concerned about reviewing every line as long as tests pass.

mnaoumov:
Closed as duplicate of #277

PR #278, opened 2022-07-12
Refactor Ignore Logic and Ignore Images for Bare URLs

### LINKED ISSUE #275: Bug: no-bare-urls adds brackets incorrectly to embedded images
## Describe the Bug

When linting, Obsidian Linter incorrectly adds brackets to embedded images in a way that breaks the image.

## How to Reproduce

``` markdown
<!--Correct: Does not add brackets-->
[text](https://www.example.com)

<!--Incorrect: Linter adds brackets-->
![image](https://www.example.com/foo.png)
<!--becomes...-->
![image](<https://www.example.com/foo.png)>
```

As shown above, the brackets are added as `(<link)>` instead of `(<link>)`, which would still work...

## Expected Behavior

``` markdown
<!--Either-->
![image](https://www.example.com/foo.png)
<!--Or-->
![image](<https://www.example.com/foo.png>)
```

## Device

- [x] Desktop
- [x] Mobile

### LINKED ISSUE #276: Refactor Idea: Move Ignore Logic to Separate File
As it stands, we have a lot of logic in `utils.ts` that is dedicated to ignoring different kinds of elements. I propose that we move this logic out to its own file either called `ignores.ts` or `ignore-types.ts`. In this file we would declare a couple of things like an enum and or type that enumerates the types we want to ignore. It would then have an associated value that is either a string, RegExp, or function depending upon the logic needed for ignoring the type specified.

This would allow us to create a generic function like so:
``` Typescript
function ignoreSpecificTypes(text: string, typesToIgnore: Array<IgnoreTypes>, func: function(text): string): string {
  if (typeToIgnore.length === 0) {
    return text;
  }

  for (const typeToIgnore of typesToIgnore) {
    if (typeof typeToIgnore === 'string') { // mdast
      getPositions(...)
     ...
    } else if (typeof typeToIgnore === 'RegExp') { // regex to find out what to ignore
      const matches = text.match();
      text = text.replace(typeToIgnore, placeholder);
    } else if (typeof typeToIgnore === 'function') { // more complex ignore logic
    let result = typeToIgnore(text);
    text = result.text;
   ...
   }
 }

  text = func(text);

  // replace placeholders with original values
  ...
  return text
}
```

This would cut down on the amount of logic in `utils.ts` and help us make ignoring different things easier than it currently is. I am working on this as a whole, but I am open to feedback.

pjkaufman:
@platers , how does the refactor for the ignore logic look? Do we want to keep the intermediary methods like `ignoreTables` and `ignoreCodeBlocksYAMLTagsAndLinks` instead of `ignoreTypes`, but use `ingoreTypes` there instead?

Fixes #276 
Fixes #275 

Refactored logic for ignoring different types of elements. It should allow for extensibility down the road as needed. Also fixed an issue where images were being caught in no bare URLs.

Changes Made:
- Removed ignore methods in `utils.ts` in favor of `ignore-types.ts` functions
- There is now an object that holds the ignore types that can be used that have their actions and placeholders defined which allows for easier maintenance of all the different types to ignore
- Standardizes how ignoring types works and allows for the ability to specify as many as you would like at once
- Updated method referenced as a utility method name since the other one was removed
- Added the ability to ignore images for no bare URLs and added associated UT

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,10 +9,10 @@ with making this plugin better.
 2. Clone the repository
 3. Run `npm ci` to install dependencies
 4. Add a new rule in `rules.ts`.
-   1. Insert a new rule in the corresponding rule type(spacing, headings, etc)
+   1. Insert a new rule in the corresponding rule type (spacing, headings, etc)
    2. Follow the format of the existing rules
    3. Add tests for edge cases in `src/test/rule_alias_here.test.ts`
-   4. You should probably use some helper functions from `utils.ts`, such as `ignoreCodeBlocksYAMLTagsAndLinks`.
+   4. You should probably use some helper functions from `utils.ts`, such as `formatYAML`.
 5. Run `npm run compile` to build, generate documentation, and test the plugin.
 6. Run `npm run lint` to lint the plugin.
 
--- a/src/ignore-types.ts
+++ b/src/ignore-types.ts
@@ -1,0 +1,147 @@
+import {getPositions, escapeDollarSigns, obsidianMultilineCommentRegex, tableRegex, tagRegex, wikiLinkRegex, yamlPlaceholder, yamlRegex} from './utils';
+import type {Position} from 'unist';
+
+
+export type IgnoreResults = {replacedValues: string[], newText: string};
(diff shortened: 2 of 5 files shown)


platers (approved):

platers:
LGTM, no need for the old methods anymore

PR #312, opened 2022-07-25
Use dedent indents consistently

### LINKED ISSUE #311: Use dedent consistently
I suggest to keep the same format for dedent usages, to visually see the text lines aligned

```javascript
obj1 = {
  obj2 = {
    key = dedent`
      line 1 indented with two spaces
      line 2 indented with two spaces

      previous line is completely empty with no spaces
    `
  }
}
```

mnaoumov:
Fixes #311


pjkaufman:
Do you mind explaining what is inconsistent about the use of dedent here?

pjkaufman:
I think I understand now. Did the linter not undo the change you are referring to?

pjkaufman:
If we are making this change, we need a way to enforce it via the linter or a config that works for several editors.

mnaoumov:
@pjkaufman no, linter doesn't touch those changes. `eslint` is not aware of `dedent`, so it can't change anything. I'm looking if we can write our own rules for `eslint`

mnaoumov:
This looks promising https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/template-indent.md

mnaoumov:
@pjkaufman I configured the linter rule to have `dedent` strings consistent. Also worked around those hard-to-catch cases with deliberate trailing spaces

CONTRIBUTING.md
```diff
@@ -126,6 +126,16 @@ Tests are run by jest. They can be run by either running `npm run test` or `npm
 
 Files should be linted prior to creating a pull request. The linter will make sure that the code follows the desired codestyle and formats the files as needed. The linter can be run by `npm run lint` or `npm run compile`
 
+Note that linter removes trailing spaces and blank lines. In case if they are essential use the following trick to preserve them
+
+```js
+const str = dedent`
```
pjkaufman, line 132: This feels very hacky.

pjkaufman (commented):
I am not sure I agree with removing blank lines or trailing space. Can we not use a setting on the rule to avoid the hacky solution?

CONTRIBUTING.md
mnaoumov: I checked the source code of the plugin

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/20e959ffadeca007697534f1ad58c45d29cfdb04/rules/template-indent.js#L72

They do unconditional trimming.

I'll send them a PR to make it configurable

But honestly invisible trailing spaces are difficult to read anyway. With this "hack" we improve code readability

mnaoumov:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1868

CONTRIBUTING.md
pjkaufman: I will take a closer look later on today.

src/rules/trailing-spaces.ts
```diff
@@ -52,11 +52,11 @@ export default class TrailingSpaces extends RuleBuilder<TrailingSpacesOptions> {
         description: 'With `Two Space Linebreak = true`',
         before: dedent`
           # H1
-          Line with trailing spaces and tabs.  
+          Line with trailing spaces and tabs.  ${''}
         `,
         after: dedent`
           # H1
-          Line with trailing spaces and tabs.  
+          Line with trailing spaces and tabs.  ${''}
```
pjkaufman, line 59: It looks like there should be a tab present here based on the text. But maybe that is incorrect.

src/rules/two-spaces-between-lines-with-content.ts
```diff
@@ -29,84 +29,84 @@ export default class TwoSpacesBetweenLinesWithContent extends RuleBuilder<TwoSpa
...
+          </div>
+          Even more content here
+          ${''}
+        `,
         after: dedent`
-            # Heading 1
-            First paragraph stays as the first paragraph
-
-            - list item 1
-            - list item 2
```
pjkaufman, line 76: Indentation was erroneously removed.

pjkaufman (approved):

src/rules/two-spaces-between-lines-with-content.ts
pjkaufman: Disregard this one.

PR #329, opened 2022-08-01
Use Obsidian Moment Instead of Local Locale Aware Moment

pjkaufman:
Using a local locale aware version of moment leads to a large increase in plugin size which leads to slower load times. If possible, it is probably better to go ahead and use the exported moment from the obsidian app as it is already locale aware. It will give improved startup times while also allowing us to use locale specific moment instances to maintain the locale specific logic in the service.

This cuts mobile loading time on my phone down from somewhere around 300 or 400 ms to about 220ms. The compiled file size is also 450ish KB instead of 750ish KB.

@mnaoumov , I wanted to see if you had any input on this change as you did a lot around getting the localization setup for moment.

Changes Made:
- Moved test folder from `src/test` to `tests` as this seems to be a more standard way of doing things and jest seems to play better with it this way
- Moved `__mocks__` into `tests` and made sure to add the export for moment which is the development instance of moment
- Updated documentation around tests directory
- Updated logic pertaining to local instance of moment to use the one provided by obsidian making sure to update the locale on it where needed


mnaoumov:
The reason for `__mocks__` folder location was from the [jest docs](https://jestjs.io/docs/manual-mocks#mocking-node-modules)](https://jestjs.io/docs/manual-mocks#mocking-node-modules)

If you override it in the `jest.config.js`, then `__mocks__` name is not longer relevant and you probably can come up with something friendlier instead

pjkaufman:
> The reason for `__mocks__` folder location was from the [jest docs](https://jestjs.io/docs/manual-mocks#mocking-node-modules)
> 
> If you override it in the `jest.config.js`, then `__mocks__` name is not longer relevant and you probably can come up with something friendlier instead

I saw that as well, but looking at a few tutorials on the subject suggested a `tests` folder which would automatically expect the mocks in that folder. I have not found documentation to support this, but it does work.

pjkaufman:
Looks like the preffered location for jest tests is `__tests__`: [jest docs](https://jestjs.io/docs/configuration#testmatch-arraystring).

So I guess that means I will move mocks back to where it was and rename tests to the default location name.

PR #336, opened 2022-08-05
Add Rule to Format Yaml Arrays

### LINKED ISSUE #172: FR: Format YAML 'tags' Key
sometimes we have

tags: tag1 tag2 tag3
Tags: tag1, tag2, tag3
tags: 
       - tag1
       - tag2
       - tag3

but would be good if they can be linted to rules, or proper yaml

tags: [tag1, tag2, tag3]

### LINKED ISSUE #182: FR: Convert Aliases to Multiline Array
The option so that you can do from this:
```
---
aliases: h1
---
```
This:
```
---
aliases: 
- h1
---
```

### LINKED ISSUE #291: FR: Format YAML Arrays
I would like a feature that unwraps yaml collections in frontmatter (per example 2.3 [here](https://yaml.org/spec/1.2.2/#21-collections)). This was an option on an older obsidian linter.

Example:

```
---
tags: [red, yellow, green]
aliases: [Note, 'Note , Comma', BigNote]
===
```

would convert to:

```
---
tags:
  - red
  - yellow
  - green
aliases:
 - Note
 - Note , Comma
 - BigNote
---
```

The [dataview plugin](https://github.com/blacksmithgu/obsidian-dataview) allows arbitrary fields in frontmatter, so this functionality would not be limited to fields recognized by Obsidian. (Perhaps there would be an option to list the fields that should be unwrapped)

I have not been able to think of any alternatives other than manually doing this.

pjkaufman:
Fixes #172 
Fixes #291
Fixes #182 

Changes Made:
- Added a rule to account for updating yaml array formats and forcing a regular string value to be a yaml array
- Added a bunch of tests to handle some possible test cases
- Added a couple of examples
- Added a fix for a bug where if a url had `[` in it, it was considered to be a regular markdown link which caused multiple `<` to be added at the start of the link when the rule for removing space around link text was active

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Documentation for all rules can be found in the [rules docs](https://github.com/
 
 - [escape-yaml-special-characters](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#escape-yaml-special-characters)
 - [format-tags-in-yaml](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#format-tags-in-yaml)
+- [format-yaml-array](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#format-yaml-array)
 - [insert-yaml-attributes](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#insert-yaml-attributes)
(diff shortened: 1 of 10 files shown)


pjkaufman:
I need to fix that it it is removing the quotation marks around single and double quotes in the yaml arrays when it adds them back.

PR #370, opened 2022-08-20
Add missing setting control

### LINKED ISSUE #369: FR: `Format Yaml Array` rule doesn't have a setting to disable formatting tags
`Format Yaml Array` rules has a setting `Format yaml aliases section` but doesn't have corresponding `Format yaml tags section`

mnaoumov:
Fixes #369


pjkaufman:
I did not realize I forgot to add the setting. Thanks for opening a PR for this.

PR #373, opened 2022-08-23
Add missing doc

### LINKED ISSUE #369: FR: `Format Yaml Array` rule doesn't have a setting to disable formatting tags

mnaoumov:
Fixes #369 

Add missing changes from #370

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -211,6 +211,8 @@ Options:
 	- `single-line space delimited`: ```tags: [tag1 tag2]```
 	- `single string space delimited`: ```tags: tag1 tag2```
 	- `single string comma delimited`: ```tags: tag1, tag2```
+- Format yaml tags section: Turns on formatting for the yaml tags section.
+	- Default: `true`
 - Default yaml array section style: The style of other yaml arrays that are not `tags`, `aliases` or  in `Force key values to be single-line arrays` and `Force key values to be multi-line arrays`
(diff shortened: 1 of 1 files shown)


docs/rules.md
```diff
@@ -211,6 +211,8 @@ Options:
 	- `single-line space delimited`: ```tags: [tag1 tag2]```
 	- `single string space delimited`: ```tags: tag1 tag2```
 	- `single string comma delimited`: ```tags: tag1, tag2```
+- Format yaml tags section: Turns on formatting for the yaml tags section.
```
platers, line 214: This file is auto generated from rules by docs.ts, you probably want to make this change somewhere else.

mnaoumov: Yes, this file is indeed autogenerated. If it is not meant to be committed, then I suggest this file to be added to the `.gitignore`

pjkaufman: @mnaoumov , the changes listed here were created by running `doc.js`  via the npm commands, right? If so, then we are good to have this merged.

I just want to clear up any questions around how this change was created.

platers: Oh sorry, confused myself. Yep, I think we should keep it this way, I don't see another way to host the docs.

PR #375, opened 2022-08-23
Add unit test to ensure all properties can be initialized

### LINKED ISSUE #374: Add an unit test to ensure that all option class properties have corresponding option control
Inspired by #369

mnaoumov:
Fixes #374

### DIFF
--- a/__tests__/setting-controls.test.ts
+++ b/__tests__/setting-controls.test.ts
@@ -1,18 +1,24 @@
 import RuleBuilder, {OptionBuilder, RuleBuilderBase} from '../src/rules/rule-builder';
 import {Options, rules} from '../src/rules';
 import '../src/rules-registry';
+import 'reflect-metadata';
 
 describe('Setting controls initialize all options class keys', () => {
   for (const rule of rules) {
     describe(rule.name, () => {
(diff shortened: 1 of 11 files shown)


pjkaufman:
I have not forgotten about this, but I do see there is reflection involved which is typically something I avoid like the plague.

mnaoumov:
@pjkaufman I can rewrite without reflection if that's your only concern :)

mnaoumov:
@pjkaufman ready

pjkaufman:
It looks good from what I have seen so far. I will take a closer look at the no settings logic and make sure I follow. Once that is done, if I see no issues, I will merge this in. Thanks for the help @mnaoumov !

mnaoumov:
@pjkaufman in short, the logic is the following: all properties from corresponding `Options` classes should have matching settings UI control, unless we explicitly marked that property with `@RuleBuilder.noSettingControl()`

pjkaufman:
Yeah. That is the gist of what I am seeing. But I want to make sure I understand the code before approving. It will likely be in the next day or 2 when I finish my review based on my freetime.

pjkaufman (approved):
Looks good to me. I will merge it on in. Thanks @mnaoumov  for adding this!

PR #390, opened 2022-09-04
Feat: custom commands

### LINKED ISSUE #358: FR: Support custom lint rules or scripts
## Problem

The current lint rules are becoming more robust as developers work on them, but they currently lack a certain amount of flexibility. For example, for the title rule in yaml, I need to read the name of the parent folder as a prefix and add it to the title. Such rules are too personal to be integrated directly into the lint plugin.

This may be more common, and it's hard for developers to add options one by one in the settings.

## Solution

This solution can be difficult or easy, depending on the developer, and I offer the following ideas for reference.

Probably the easiest：
Add a new lint rule and execute the script after the other rules have run. This part can even be further simplified with quickadd, which executes the quickadd command after the other lint rules are executed.

Probably the most difficult:
Similar to eslint, standardize, register and manage lint rules, load into existing lint rules via npm packages

image

## Additional Context

How do you think this feature should be handled better? Looking forward to any response from you.

windily-cloud:
Fixes #358 

I added a setting item about custom commands, and the user sets the commands to be stored in the `lintCommands` array in data.json, and I intend to execute the commands after the lint rules are executed (this step has not been added yet). 

If there is anything that needs to be improved, please feel free to mention it. I am not very familiar with eslint rules and ts.


src/main.ts
```diff
@@ -106,7 +107,7 @@ export default class LinterPlugin extends Plugin {
         const startMessage = 'This will edit all of your files and may introduce errors.';
         const submitBtnText = 'Lint All';
         const submitBtnNoticeText = 'Linting all files...';
-        new LintConfirmationModal(this.app, startMessage, submitBtnText, submitBtnNoticeText, () =>{
+        new LintConfirmationModal(this.app, startMessage, submitBtnText, submitBtnNoticeText, () => {
```
pjkaufman, line 110: Are you using the eslint rules to make these changes or is this your editor/IDE that is making the spacing changes? I am curious since this will just be undone if it is not being created by eslint.

src/main.ts
```diff
@@ -553,6 +557,71 @@ class SettingTab extends PluginSettingTab {
...
+      const element = arr[fromIndex];
+      arr[fromIndex] = arr[toIndex];
+      arr[toIndex] = element;
+    }
+    new Setting(this.containerEl)
+        .addButton((cb)=>{
+          cb.setButtonText('Add new command')
+              .setCta()
+              .onClick(()=>{
+                this.plugin.settings.lintCommands.push('');
```
pjkaufman, line 576: You seem to be pushing an empty value instead of the selected value, but maybe that was something you planned to add yet.

src/main.ts
```diff
@@ -553,6 +557,71 @@ class SettingTab extends PluginSettingTab {
...
+
+    this.plugin.settings.lintCommands.forEach((command, index) => {
+      new Setting(this.containerEl)
+          .addSearch((cb) => {
+            new CommandSuggester(this.app, cb.inputEl);
+            cb.setPlaceholder('Example: folder1/template_file')
+                .setValue(command)
+                .onChange((newCommand) => {
+                  if (newCommand && this.plugin.settings.lintCommands.contains(newCommand)) {
+                    logError('This command is already exist!', new Error('This command is already exist!'));
```
pjkaufman, line 590: The following is a minor grammar fix and a change in wording in the log to make it clearer and not repeat the same info twice:
```suggestion
                    logError('You cannot add the same command to the list of custom rules twice.', new Error('this command already exists!'));
```

src/suggesters/command-suggester.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(
+    public app: App,
+    public inputEl: HTMLInputElement,
+  ) {
+    super(app, inputEl);
+  }
+
+  getSuggestions(input_str: string): Command[] {
+    // @ts-ignore
+    const all_commands = this.app.commands.listCommands();
```
pjkaufman, line 14: Why do we need `@ts-ignore` here? It looks like we could just add the missing type info to the module declaration in  `main.ts`: https://github.com/platers/obsidian-linter/blob/master/src/main.ts#L32-L40

So something like the following (my syntax for method declarations in an interface may be wrong):
``` Typescript
declare module 'obsidian' {
  // eslint-disable-next-line no-unused-vars
  interface App {
    commands: {
        executeCommandById(id: string): void;
        commands: {
            'editor:save-file': {
                callback(): void;
            };
        listCommands: function(): Command[],
        };
    };
  }
}
```

pjkaufman (commented):
I am not sure if these were files copied from another repo. If they were, we should give credit to them like what was done in `src/suggesters/suggest.ts`. Beyond that, I have some relatively minor suggestions and questions. However I am not really sure that using Obsidian commands is the right way to go about this as it may make us rely on other plugins to run special lint commands. While this is not necessarily bad, it makes this a very difficult feature to make sure is working which makes it harder to maintain the plugin itself. I am not entirely opposed to this, but we do have to see if there are any ways we can get around the issues of not being able to easily verify that the feature works.

src/main.ts
windily-cloud: This is the change I made when I ran `npm run lint`, and I don't know why it's different from yours.

src/main.ts
windily-cloud: Yes, this initializes an empty string when the command is added, and is later changed by the user.

src/suggesters/command-suggester.ts
windily-cloud: You are right!

windily-cloud: But I have doubts here, directly import type from obsidian.d.ts, maybe need not to declare an obsidian package?

src/main.ts
pjkaufman: Gotcha. It could be that it is doing something weird on my OS. But this should be fine then.

src/main.ts
pjkaufman: Interesting. It might help to add a comment saying that is the case to make it clear what is happening as I guess I just misunderstood it.

src/suggesters/command-suggester.ts
pjkaufman: Some properties on the actual interface are not exported. This seems to be one of them. Thus we have to add a local module declaration to add in the missing properties. It is generally only meant for use in cases where you would otherwise need to do a type cast or `@ts-ignore` to get eslint to know it exits.

windily-cloud:
> I am not sure if these were files copied from another repo. If they were, we should give credit to them like what was done in `src/suggesters/suggest.ts`. 

I had write a declare in the header
// Credits go to Liam's Periodic Notes Plugin: https://github.com/liamcain/obsidian-periodic-notes
image

>  but we do have to see if there are any ways we can get around the issues of not being able to easily verify that the feature works.

It may be difficult for me to write such a relatively complex function by myself, relatively speaking, this is the best way I can think of, I also rely heavily on plugins like quickadd, the author is also very nice. If you have a better idea or need help, I'll try my best

pjkaufman:
> > I am not sure if these were files copied from another repo. If they were, we should give credit to them like what was done in `src/suggesters/suggest.ts`.
> 
> I had write a declare in the header // Credits go to Liam's Periodic Notes Plugin: https://github.com/liamcain/obsidian-periodic-notes image
> 
> > but we do have to see if there are any ways we can get around the issues of not being able to easily verify that the feature works.
> 
> It may be difficult for me to write such a relatively complex function by myself, relatively speaking, this is the best way I can think of, I also rely heavily on plugins like quickadd, the author is also very nice. If you have a better idea or need help, I'll try my best

I can handle the UTs as I believe I know how to mock the Obsidian Command as I have done something similar already.

I think that what is here is fine for the actual settings page. It would just need to have the actually running of the commands added. I am not sure if you plan on adding that or if you would like me to try to bring this to fruition.

My initial concern was arouns requiring Quick Add to add custom commands. But I see that we are not doing that.

windily-cloud:
As long as the linter plugin provides the ability to run a command during lint I'm satisfied, the rest of the features and optimizations I don't require. You can do anything you want with the code I wrote.

pjkaufman:
By using the changes made, I have found a couple of issues that are minor that I am working on.

On load, there is nothing that adds the existing settings to the UI which means on each load the list of custom commands is empty.

When you select the same option as the currently selected option in the same command you get an error despite the fact that there is only one instance of the rule (it thinks you are adding a new one instead of knowing that the one modified and the one it is matching with are the same).

I am working on getting it to a testable state which will hopefully be soon, but we will see about that.

pjkaufman:
There is also an issue where when you add a new command the cursor is not moved to the new command search box which can cause visual issues and confusion because it will not display the newly created command search box as you need to scroll to see it.

pjkaufman:
I was able to add a filter to make sure that the already excluded linter values do not show up in the list. However we may want to consider only counting it as a save when you either click or hit enter as it seems to trigger a save when you type in a text change. This can erroneously fire the check for unique command values.

I also made the search include the name of the id. I think we should save both the command id and command name for the suggested value since we will likely want to display the command name while we use the command id to execute the command behind the scenes.

pjkaufman:
@windily-cloud , the code changes I have made thus far get the logic to a testable state. It goes beyond just what was necessary to add the feature. It still needs the UTs added, but it should not be too bad to do that today.

Here is what I did:
- Moved the logic for running the linter to its own class
- Reworked error handling to better allow for showing issue with rules (still need to add that for custom lint commands)
- Made sure the loaded config value would set the value on the setting
- Added a custom lint error to help identify when we know which rule had the problem
- Updated typing location to make sure it would not cause a circular import dependency
- Updated the suggester search to include the name of commands to make sure it would allow easier command searching

What still needs to happen:
- When add command is selected, move cursor to new command search box
- When editing the current search value and hitting enter, make sure it does not say the setting already exists
- See about making the value for the lint commands both the name and id so it can have a human friendly display
- See about updating the warning about not using the same command twice to be something that displays to the user
- UTs for running the tests should have their own UTs (this will take a little while to write out, but should give some confidence that it works)

pjkaufman:
I have an update that should take care of displaying the name of the command instead of the command id while retaining info on both of them. I believe I also have a fix for the same command showing up twice being an issue. But it seems to create another issue where clearing out the value in a command does not let you search for that exact same command.

I see that on reload we are going ahead and refreshing the settings tab which will make it difficult to just go ahead and set the value of where the cursor will go. So for now, I will not touch that.

pjkaufman:
I realized that without actually being able to emulate the obsidian API fully, all we can really test is that the logic gets called correctly. There are more tests that should be added when it comes to disabling rules and things like that, but that is beyond the scope of this PR.

This just leaves the two remaining things that I would like to fix up. It may take me a while to figure out the warnings as I am not a front end dev and React is newer to me. The issue with moving to the latest addition of a search box might not be something that can be easily fixed. I had an idea, but without knowing how to properly identify elements in React, I was not really able to make much progress.

windily-cloud:
>  When add command is selected, move cursor to new command search box

I added this feature, but I don't know if it's what you're thinking of:
- Click `Add new command` button，then the the last input will be focused.
- In a input box, click tab, the next input will be focused.

windily-cloud:
>  See about updating the warning about not using the same command twice to be something that displays to the user

I don't really understand what this means. We just need to restrict the user from entering the saved commands. In my opinion there should be no need to run the same command multiple times.

pjkaufman:
> > See about updating the warning about not using the same command twice to be something that displays to the user
> 
> I don't really understand what this means. We just need to restrict the user from entering the saved commands. In my opinion there should be no need to run the same command multiple times.

Thanks for adding the focus. I think that is it (I am not in the front end too frequently, so I often fofget the terms for these things).

I was thinking we add a little html paragraph right below each search element. When a user updates the contents of the search, but does not select a command, the value is not saved, so we should display yellow text explaining that the value has not saved.

Another case is to display red text in that paragraph when they somehow select the same command twice (right now that should not be feasible and we might just be able to move that logic to the backend where we run the actual custom commands).

windily-cloud:
> I was thinking we add a little html paragraph right below each search element. 

I don't think it's necessary to do so, because the current practice is the common logic in obsidian, quickadd does it, templater does it, and almost all other plugins do it.

> Another case is to display red text in that paragraph when they somehow select the same command twice

The input box has an onchange event, which I initially thought would solve the problem if we checked in there if it contained an existing command.

Didn't you already add a filter to it? I think that's enough for now.

pjkaufman:
That worls for me. I can throw the error. The only thing I will see about doing is adding a description for custom lint commands.

I think the settings may need to move to a tabbed layout to help make rule navigation easier. But that would come later.

windily-cloud:
I've only learned a bit of programming since I've been using obsidian(programming for Google ...), and I've realized that writing the UI in settings with native descriptors is very cumbersome and testing is a pain. If I could import react or vue it would probably be much better, but the linter plugin obviously doesn't need to do that, it's configured once and can be used all the time without frequent configuration changes.

I strongly support you to make linter more user-friendly.

pjkaufman:
The only thing that I am thinking still needs to happen is the addition of documentation on how to create a macro that can be run by the user. This should not be too bad. Once that is present, I am fine with going ahead and merging this. I may not move where the logic is stored and possibly setup a place for UI specific logic, but that would come down the road.

pjkaufman:
@windily-cloud , please take a look and let me know if you notice any issues with the functionality or the UI. If you see no issues, I believe this is ready to merge into master to fix #358 .

windily-cloud:
Awesome!!! I have tried the feature several times, no problems were found. Thanks for your efforts to make the linter plugin even more powerful.👍👍👍👍

PR #406, opened 2022-09-12
RemoveEmptyListMarkers fix for single number

### LINKED ISSUE #405: Bug: `Remove Empty List Markers` rule removes number if it's the only content in the line
## Describe the Bug

`Remove Empty List Markers` rule removes number if it's the only content in the line

## How to Reproduce

Consider note

``` markdown
# World War II

1939
```

And run  `Remove Empty List Markers` rule.

## Expected Behavior

The note is not changed

## Actual Behavior

`1939` is trimmed from the note as it was mistakenly considered to be a list marker, which is weird, because it is not followed by a dot `.`

mnaoumov:
Fixes #405

### DIFF
--- a/src/rules/remove-empty-list-markers.ts
+++ b/src/rules/remove-empty-list-markers.ts
@@ -22,7 +22,7 @@ export default class RemoveEmptyListMarkers extends RuleBuilder<RemoveEmptyListM
   }
   apply(text: string, options: RemoveEmptyListMarkersOptions): string {
     return ignoreListOfTypes([IgnoreTypes.code, IgnoreTypes.yaml, IgnoreTypes.link, IgnoreTypes.wikiLink, IgnoreTypes.tag], text, (text) => {
-      const emptyListMarkerRegex = /^\s*(>\s*)*(-|\*|\+|\d+\D|- (\[( |x)\]))\s*?$/gm;
+      const emptyListMarkerRegex = /^\s*(>\s*)*(-|\*|\+|\d+[.)]|- (\[( |x)\]))\s*?$/gm;
       // remove all empty list markers followed by a new line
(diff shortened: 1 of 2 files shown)


src/rules/remove-empty-list-markers.ts
```diff
@@ -22,7 +22,7 @@ export default class RemoveEmptyListMarkers extends RuleBuilder<RemoveEmptyListM
   }
   apply(text: string, options: RemoveEmptyListMarkersOptions): string {
     return ignoreListOfTypes([IgnoreTypes.code, IgnoreTypes.yaml, IgnoreTypes.link, IgnoreTypes.wikiLink, IgnoreTypes.tag], text, (text) => {
-      const emptyListMarkerRegex = /^\s*(>\s*)*(-|\*|\+|\d+.|- (\[( |x)\]))\s*?$/gm;
+      const emptyListMarkerRegex = /^\s*(>\s*)*(-|\*|\+|\d+\D|- (\[( |x)\]))\s*?$/gm;
```
pjkaufman, line 25: My sole concern is that this should be `\.`. I guess I messed that before.

pjkaufman (changes requested):

src/rules/remove-empty-list-markers.ts
mnaoumov: We could also allow `)` as it is supported by [CommonMark](https://commonmark.org/help/)

pjkaufman: I am fine with that as well. It would then just need an or in the regex to check either `.` or `)`. So the following should do the trick: `(\.|\))`.

pjkaufman (approved):
Looks great! Thanks for reporting and creating a fix, @mnaoumov !

PR #414, opened 2022-09-17
Format Ordered Lists

### LINKED ISSUE #153: FR: Allow for Automatically Numbering Ordered Lists
Screenshot 2021-12-28 00 33 20

When editing numbered lists, the numbering is often off. It would be great to autofix this. Afair, markdown linters have this, so probably this isn't something we would have to implement ourselves?

pjkaufman:
Fixes #153 

Changes Made:
- Made sure that you can update the number value of ordered lists to either be ascending or to be `1`.
- Made sure that you can have the end of the ordered list indicator/marker be either `.` or `)`
- Added examples including nesting in blockquotes


seyeeL:
image
Sub-levels not in effect

pjkaufman:
> image Sub-levels not in effect

This looks like a bug. Could you open up a bug and include the text rather than an image? I think I know how to reproduce this, but it will take me a minute to think through how to properly determine when it is no longer the same sublist.

Thanks for pointing this out!

seyeeL:
https://github.com/platers/obsidian-linter/issues/422#issue-1378700510

PR #433, opened 2022-09-30
Fix empty aliases section usecase

### LINKED ISSUE #432: Bug: `YAML Title Alias` rule doesn't respect `Keep alias that matches the filename` setting when aliases section is empty
## Describe the Bug

`YAML Title Alias` rule doesn't respect `Keep alias that matches the filename` setting when aliases section is empty

## How to Reproduce

Consider

**Filename.md**

``` markdown
---
aliases:
---
# Filename
```

Configure `YAML Title Alias` rule setting `Keep alias that matches the filename` = `false`. Run the rule.

## Expected Behavior

`aliases` section has to removed

``` markdown
---
---
# Filename
```

## Actual Behavior

`aliases` section adds `Filename` alias ignoring the setting mentioned.

``` markdown
---
aliases:
  - Filename
---
# Filename
```

mnaoumov:
Fixes #432

### DIFF
--- a/__tests__/yaml-title-alias.test.ts
+++ b/__tests__/yaml-title-alias.test.ts
@@ -1000,5 +1000,22 @@ ruleTest({
         preserveExistingAliasesSectionStyle: false,
       },
     },
+    {
+      testName: 'Removes empty alias section if title matches the filename',
+      before: dedent`
+        ---
+        aliases:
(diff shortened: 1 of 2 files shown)


pjkaufman (approved):
Thanks for catching this! It looks good.

PR #455, opened 2022-10-21
Added a function for custom regex replacement

HardwayLinka:
Hi, I added a new command
image

we can add that command to the Custom Commands
image

Finally, we are able to use custom regular expressions to replace text
regex

I think this is a useful requirement, hope it helps others


pjkaufman:
Hey @HardwayLinka . This is a good suggestion. I think it can be really useful.

The only problem I have with it is that it does not really fit under the concept of custom commands in the sense that it is a command you can run without specifying params. Thus it seems like it would be better as its own option of custom rules called custom replacement or something like that. Thus it would not need to be added onto custom commands. I am just not sure how well that would work with the command form of it. I would likely need to refactor it some so that it could both run as a rule and as a command.

What fo you think of my feedback? I would be happy to help implement changes where needed.

HardwayLinka:
Hi @pjkaufman.thank you for your patient and detailed feedback.

I agree with what you said. I added this feature to the custom command because it is easier to implement, but at the same time it may be against your design. I will try adding this to a separate option and commit it again when I'm done.

HardwayLinka:
Hi @pjkaufman. I have a question, I added the feature to the Content Settings, but I can't find the feature under the Content Settings

image

pjkaufman:
> Hi @pjkaufman. I have a question, I added the feature to the Content Settings, but I can't find the feature under the Content Settings
> 
> image
> 
> image

Hey @HardwayLinka , I see that you added it as a regular rule. It would need to be a custom  rule just like custom commands (just different UI elements).

As for it not showing up, I am guessing you did a search for Content instead of the name of the rule or part of the description, or part of the rule alias. That would be why it did not show up though I might be misunderstanding  the screenshot .

src/main.ts
```diff
@@ -370,6 +374,8 @@ export default class LinterPlugin extends Plugin {
     } catch (error) {
       this.handleLintError(file, error, 'Lint File Error in File \'${file.path}\'', false);
     }
+
+    this.rulesRunner.runCustomRegexReplacement(this.settings.customRegexs, editor);
```
pjkaufman, line 378: I would actually have the custom regex rules run at the end of the regular rules in the rulesRunner when a file gets linted.

src/main.ts
```diff
@@ -221,6 +222,9 @@ export default class LinterPlugin extends Plugin {
     if (Object.prototype.hasOwnProperty.call(storedSettings, 'lintCommands')) {
       this.settings.lintCommands = storedSettings.lintCommands;
     }
+    if (Object.prototype.hasOwnProperty.call(storedSettings, 'customRegexs')) {
+      this.settings.customRegexs = storedSettings.customRegexs;
+    }
```
pjkaufman, line 227: This may need updating to the new form of settings that just got pushed up, but if not, I can go ahead and update it after the fact.

src/ui/settings.ts
```diff
@@ -59,7 +59,10 @@ export class SettingTab extends PluginSettingTab {
       this.addRuleToTab(tabTitle, rule);
     }
 
-    this.createTabAndContent('Custom', navEl, settingsEl, (el: HTMLElement, tabName: string) => this.generateCustomCommandSettings(tabName, el));
+    this.createTabAndContent('Custom', navEl, settingsEl, (el: HTMLElement, tabName: string) => {
+      this.generateCustomCommandSettings(tabName, el);
+      this.generateCustomRegexReplacementSettings(tabName, el);
```
pjkaufman, line 64: These probably need to be added into their own divs so the search for them will not work in a wonky manner.

src/ui/settings.ts
```diff
@@ -238,6 +241,54 @@ export class SettingTab extends PluginSettingTab {
     });
   }
 
+  generateCustomRegexReplacementSettings(tabName: string, containerEl: HTMLElement): void {
```
pjkaufman, line 244: This is mostly good to go. It just looks like it is missing in the list of rules to search which requires a container el that is not the direct container el for another setting, the tab name, and a set of information to search on .

pjkaufman (changes requested):
Overall this looks pretty good, but there are a couple of things I pointed out on the PR. Overall it is looking great!

src/main.ts
HardwayLinka: Hey, I am confused where is the new form of settings

pjkaufman: Here is the new format they follow: https://marcus.se.net/obsidian-plugin-docs/user-interface/settings

It uses default settings in main: https://github.com/platers/obsidian-linter/blob/master/src/main.ts#L43-L58

src/ui/settings.ts
```diff
@@ -60,6 +60,7 @@ export class SettingTab extends PluginSettingTab {
     }
 
     this.createTabAndContent('Custom', navEl, settingsEl, (el: HTMLElement, tabName: string) => this.generateCustomCommandSettings(tabName, el));
+    this.createTabAndContent('Custom', navEl, settingsEl, (el: HTMLElement, tabName: string) => this.generateCustomRegexReplacementSettings(tabName, el));
```
pjkaufman, line 63: This change here will cause problems. Try something more like:
```suggestion
    this.createTabAndContent('Custom', navEl, settingsEl, (el: HTMLElement, tabName: string) => {
      let tempContainer = el.createDiv();
      this.generateCustomCommandSettings(tabName, tempContainer);
      tempContainer = el.createDiv();
      this.generateCustomRegexReplacementSettings(tabName, tempContainer)
    });
```

Note that this is more of sudo code and will need to be verified to see if it works in its suggested state.

src/ui/settings.ts
```diff
@@ -238,6 +239,59 @@ export class SettingTab extends PluginSettingTab {
...
+                .setTooltip('Delete')
+                .onClick(()=>{
+                  this.plugin.settings.customRegexs.splice(index, 1);
+                  this.plugin.saveSettings();
+                  this.display();
+                });
+          });
+    });
+
+    this.addSettingToMasterSettingsList(tabName, containerEl as HTMLDivElement, tabName.toLowerCase(), searchInfo);
```
pjkaufman, line 292: The search info is actually the name and description of the rule itself. So you would likely just have the name of the setting placed here for search info: "Custom Regex Replacement".
A description is welcome as well though.

pjkaufman (changes requested):

src/main.ts
```diff
@@ -362,6 +363,13 @@ export default class LinterPlugin extends Plugin {
     } catch (error) {
       this.handleLintError(file, error, 'Lint File Error in File \'${file.path}\'', false);
     }
+
+    oldText = editor.getValue();
+    newText = this.rulesRunner.runCustomRegexReplacement(this.settings.customRegexs, oldText);
+    const lines = oldText.split('\n');
+    const start = {line: 0, ch: 0};
+    const end = {line: lines.length - 1, ch: lines[lines.length - 1].length};
+    editor.replaceRange(newText, start, end, oldText);
```
pjkaufman, line 372: I am not sure this needs to be its own thing. We should be able to call `runCustomRegexReplacement` from the `lintText` method and not have to worry about recreating the logic used for running these rules for a file.

src/rules-runner.ts
```diff
@@ -130,6 +130,19 @@ export class RulesRunner {
...
   }
 
+  runCustomRegexReplacement(customRegexs: CustomRegex[], oldText: string): string {
+    logDebug(`Running Custom Regex`);
+    let tempOldText = oldText;
+    for (const eachRegex of customRegexs) {
+      if (!eachRegex.find || eachRegex.find.trim() == '' || !eachRegex.replace || eachRegex.replace.trim() == '') {
+        continue;
+      }
+      const regex = new RegExp(`${eachRegex.find}`, 'gm');
```
pjkaufman, line 140: Is there a reason not to allow for the user to select the regex flags? I know with custom ignore, I am allowing them to select which flags to use (I am not going to hold this getting merged in over that as I will likely need to update the display to allow for those selections when I finish up custom ignore).

pjkaufman (changes requested):
Minus the one thing about the order in which this runs, it looks great. A benefit to running the custom regexes after the regular linting rules, but before the final linting rules (i.e. [here](https://github.com/platers/obsidian-linter/blob/master/src/rules-runner.ts#L65) that is to say before line 66) is that it will not cause issues with date modified and other things when running the custom regex.

pjkaufman:
Also, sorry about the delay in responding. I was out over the weekend. Thanks for putting in the effort to add this feature!

HardwayLinka:
There's no need to apologize. This project has been very helpful to me, so I'd love to contribute to this project and hope it will help others~

pjkaufman:
I will take a look at this hopefully some time this week and then determine if any changes need to be made just from a trail run of it. As far as the code looks, it looks good. I just want to make sure I did not overlook something that would cause an odd user experience.

pjkaufman:
This looks good. It looks to cover several different cases as well. I will go ahead and merge this so we can get it out there in the next release. I will also probably add some documentation around this in the README and refactor this and the custom commands into their own components instead of as they are now. Sorry about the delay on getting to this to verify that it looked to work. 

Thanks for putting in the work to get this working! I am sure there will be people that find this useful.

pjkaufman (approved):

PR #457, opened 2022-10-23
Fix look of inactive settings tab icons

chrisgrieser:
the name of the inactive tab is not properly hidden, causing issues with some themes like mine. This PR adds a rule which should properly fix this issue for any theme.

without the PR:
Pasted image 2022-10-23 19 46 03

with the PR:
Pasted image 2022-10-23 19 47 43

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -64,6 +64,10 @@
         max-width 0.45s cubic-bezier(0.57, 0.04, 0.58, 1) 0.2s;
 }
 
+.linter-navigation-item:not(.linter-navigation-item-selected) > span:nth-child(2) {
+	display: none;
+}
+
 /**
  * Based on https://github.com/phibr0/obsidian-commander/blob/main/src/styles.scss
  */
--- a/src/styles.css
+++ b/src/styles.css
@@ -64,10 +64,6 @@
         max-width 0.45s cubic-bezier(0.57, 0.04, 0.58, 1) 0.2s;
 }
 
-.linter-navigation-item:not(.linter-navigation-item-selected) > span:nth-child(2) {
-		opacity: 0;
-}
-
 /**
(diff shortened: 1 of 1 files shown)


pjkaufman:
@chrisgrieser, thanks for pointing this out! Is there a reason to not use the hidden class? Display none works, but is not accessibility  friendly.

chrisgrieser:
What do you mean by "hidden class"? If you mean `visibility: hidden;`, yeah that does work, too, I didn't know that this makes a difference accessibility wise.

pjkaufman:
> What do you mean by "hidden class"? If you mean `visibility: hidden;`, yeah that does work, too, I didn't know that this makes a difference accessibility wise. Have changed the PR accordingly.

There is a style called hidden  which is not perfect, but it should be more accessibility  friendly.

pjkaufman:
I had the name wrong. Here is the style: https://github.com/platers/obsidian-linter/blob/3668268a3ca3501c699fe44ac264838734c2426a/src/styles.css#L135

chrisgrieser:
well, since the inactive tab names do not get that hidden class, I assume that would have to be done via js, and not via css, and therefore outside my area of expertise 😝 

although if I understand the article correctly, `opacity: 0` would also work to hide it while keeping it there fore screenreaders etc.

pjkaufman:
I could be wrong, but why would it not be feasible to extend visually hidden with the css selector you had? It should be able to apply to both based on what you are saying.

chrisgrieser:
to my knowledge, there is no way in (pure) css to extend the styling of one class to another?

chrisgrieser:
ah, now I understand what you mean. Yeah, adding the selector to that block also works, sure. done so in the latest commit

pjkaufman:
That's what I was going for. Thanks!

PR #465, opened 2022-10-30
[fix] hotkey not work on macOS

rea1shane:
`l` not work on Apple M1 Pro with Ventura 13.0, modify `l` to `L`.


pjkaufman:
Thanks for making that update. The changes seems to work even for Linux, so I will go ahead and approve merging this in. Hopefully it will not cause any issues.

Issue #479, opened 2022-11-06
FR: space between Chinese/Japanese/Korean and English


iucario:
Examples:

```markdown
日本語englishひらがな
カタカナenglishカタカナ
ﾊﾝｶｸｶﾀｶﾅenglish１２３全角数字
한글english한글
```

```markdown
日本語 english ひらがな
カタカナ english カタカナ
ﾊﾝｶｸｶﾀｶﾅ english１２３全角数字
한글 english 한글
```

The original rule only considers Chinese.
https://github.com/platers/obsidian-linter/blob/5624359551750c9adbb7ad67745a69677cf3259d/src/rules/space-between-chinese-and-english-or-numbers.ts#L25
To deal with CJK, we can easily implement this by replacing `[\u4e00-\u9fa5]` with `\p{sc=Han}|\p{sc=Katakana}|\p{sc=Hiragana}|\p{sc=Hangul}`

pjkaufman:
Thanks for this suggestion. I will have to see about adding this to the list of things to do. I am definitely interested in adding it, I just need to see about making sure that it works gets the right updates to map the property since it will need that to happen due to wording changes that will be necessary.

pjkaufman:
There may be a problem adding this because of the Unicode property usage, but we can go ahead and see if that truly is the case once I get through some of the things I am working on.

pjkaufman:
This change has been merged and should be present on master and in the next release. Please let us know if either of these is not the case.

PR #481, opened 2022-11-08
feat: space between CJK and English

iucario:
#479 

The option name also needs to be modified.

What do you think?

### DIFF
--- a/src/rules/space-between-chinese-japanese-or-korean-and-english-or-numbers.ts
+++ b/src/rules/space-between-chinese-japanese-or-korean-and-english-or-numbers.ts
@@ -4,31 +4,34 @@ import dedent from 'ts-dedent';
 import {ignoreListOfTypes, IgnoreTypes} from '../utils/ignore-types';
 import {updateBoldText, updateItalicsText} from '../utils/mdast';
 
-class SpaceBetweenChineseAndEnglishOrNumbersOptions implements Options {}
+class SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions
+implements Options {}
 
 @RuleBuilder.register
-export default class SpaceBetweenChineseAndEnglishOrNumbers extends RuleBuilder<SpaceBetweenChineseAndEnglishOrNumbersOptions> {
-  get OptionsClass(): new () => SpaceBetweenChineseAndEnglishOrNumbersOptions {
-    return SpaceBetweenChineseAndEnglishOrNumbersOptions;
+export default class SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbers extends RuleBuilder<SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions> {
+  get OptionsClass(): new () => SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions {
+    return SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions;
   }
   get name(): string {
-    return 'Space between Chinese and English or numbers';
+    return 'Space between Chinese Japanese or Korean and English or numbers';
   }
   get description(): string {
-    return 'Ensures that Chinese and English or numbers are separated by a single space. Follows these [guidelines](https://github.com/sparanoid/chinese-copywriting-guidelines)';
+    return 'Ensures that Chinese, Japanese, or Korean and English or numbers are separated by a single space. Follows these [guidelines](https://github.com/sparanoid/chinese-copywriting-guidelines)';
   }
   get type(): RuleType {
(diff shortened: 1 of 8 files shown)


pjkaufman:
The code change looks fine. The version change is ok, but should be reserved for the creation of a release.

The name change needs to happen, but that will require mapping from the old rule name to the new one.

pjkaufman:
Never mind my comment about version changes. It looks like that was done automatically.

iucario:
Currently the rule name is "Space between Chinese and English or numbers". What name should we rename it to? "Space between CJK and English or numbers"?

pjkaufman:
I would rename it to `Space between Chinese, Japanese, or Korean and English or numbers`. This will also require renaming the class and its file and test files. We will also need to update the description of the rule to match the changes.

We will also need to map the setting data similar to https://github.com/platers/obsidian-linter/blob/master/src/main.ts#L101-L113.

pjkaufman:
Once the file name is updated, the import for the test needs to be updated as well.

pjkaufman:
As for mapping the data, that is to preserve the current settings to keep it from accidentally losing people's settings. It is currently pretty brittle and I am working on thinking up a better way of storing settings so that they are not brittle and will better withstand wording changes and name changes.

pjkaufman:
As for how to map it, it depends entirely on what changes: a rule name change requires mapping the old rule name to the new rule name. Changing a description, requires mapping the old description to the new description.

Here is an example: https://github.com/platers/obsidian-linter/blob/master/src/main.ts#L101-L113

If you would like, I can add the settings mapping.

iucario:
`npx jest --clearCache` worked! Thanks.

I changed the file name to `space-between-chinese-japanese-or-korean-and-english-or-numbers.ts`. Please help adding the settings mapping.

pjkaufman:
I have tried to update your branch with the necessary changes for the settings mapping and one other issue I noticed, but it seems I do not have permission to do so.

You will need to add the following [here](https://github.com/iucario/obsidian-linter/blob/space-cjk/src/main.ts#L124):
``` TypeScript
const spaceBetweenChineseAndEnglishOrNumbers = this.settings.ruleConfigs['Space between Chinese and English or numbers'];
    if (spaceBetweenChineseAndEnglishOrNumbers) {
      const enabled = spaceBetweenChineseAndEnglishOrNumbers['Ensures that Chinese and English or numbers are separated by a single space. Follows these [guidelines](https://github.com/sparanoid/chinese-copywriting-guidelines)'];
      this.settings.ruleConfigs['Space between Chinese, Japanese, or Korean and English or numbers'] = {
        'Ensures that Chinese, Japanese, or Korean and English or numbers are separated by a single space. Follows these [guidelines](https://github.com/sparanoid/chinese-copywriting-guidelines)': enabled,
      };
      delete this.settings.ruleConfigs['Space between Chinese and English or numbers'];
    }
```

src/rules/space-between-chinese-japanese-or-korean-and-english-or-numbers.ts
```diff
@@ -4,39 +4,68 @@ import dedent from 'ts-dedent';
...
-export default class SpaceBetweenChineseAndEnglishOrNumbers extends RuleBuilder<SpaceBetweenChineseAndEnglishOrNumbersOptions> {
-  get OptionsClass(): new () => SpaceBetweenChineseAndEnglishOrNumbersOptions {
-    return SpaceBetweenChineseAndEnglishOrNumbersOptions;
+export default class SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbers extends RuleBuilder<SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions> {
+  get OptionsClass(): new () => SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions {
+    return SpaceBetweenChineseJapaneseOrKoreanAndEnglishOrNumbersOptions;
   }
   get name(): string {
-    return 'Space between Chinese and English or numbers';
+    return 'Space between Chinese Japanese or Korean and English or numbers';
```
pjkaufman, line 16: A grammar change:
```suggestion
    return 'Space between Chinese, Japanese, or Korean and English or numbers';
```

pjkaufman (changes requested):
Other than the addition of the change to main and this suggestion here, it looks pretty good. You will likely need to do a rebase of this branch off of your master branch once you sync it with the plugin repo. That should fix the merge conflict as well.

Sorry about the delay. Thanks for putting in the work to get this added.

pjkaufman:
Well, it looks like I was able to bypass my issue. I accidentally was trying to create a new branch instead of using the existing one. I will see about rebasing this branch and see if it fixes things.

pjkaufman (approved):

PR #548, opened 2022-12-09
feat: lint command that respects ignored folders

### LINKED ISSUE #546: FR: Add command that respects `Lint on save` setting
I am currently triggering the "Lint the current file" command using another plugin, but I would like the linting step to respect ignored folders. This requires adding a new command ("Lint the current file unless exempt") which could be used. This is essentially the same as "Lint the current file" except it respects the ignored folders.

Would you @pjkaufman take a PR that implements this?

JeppeKlitgaard:
Fixes #546. See discussion there.

Additionally contains small change in package-lock, which was not updated for `1.9.1`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-linter",
-  "version": "1.7.1",
+  "version": "1.9.1",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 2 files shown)


pjkaufman:
I don't have a problem with this. It looks good.

PR #552, opened 2022-12-13
Don't include whitespace when parsing tags

### LINKED ISSUE #551: Bug: `YAML Title` rule incorrectly considers `#tag` after header as a part of the title
## Describe the Bug

`YAML Title` rule incorrectly considers `#tag` after header as a part of the title

## How to Reproduce

Markup

``` markdown
# title
#tag
```

`YAML Title` rule.

## Expected Behavior

``` markdown
---
title: title
---
# title
#tag
```

## Actual Behavior

``` markdown
---
title: title
#tag
---
# title
#tag
```

mnaoumov:
Fixes #551

### DIFF
--- a/src/utils/regex.ts
+++ b/src/utils/regex.ts
@@ -13,8 +13,7 @@ export const codeBlockRegex = new RegExp(`${backtickBlockRegexTemplate}|${tildeB
 export const wikiLinkRegex = /(!?)\[{2}([^\][\n|]+)(\|([^\][\n|]+))?\]{2}/g;
 // based on https://davidwells.io/snippets/regex-match-markdown-links
 export const genericLinkRegex = /(!?)\[([^[]*)\](\(.*\))/g;
-export const tagRegex = /(?<=\s|^)#[^\s#;.,><?!=+]+/g;
-export const tagWithLeadingWhitespaceRegex = /(?:\s|^)#[^\s#;.,><?!=+]+/g;
+export const tagWithLeadingWhitespaceRegex = /(\s|^)(#[^\s#;.,><?!=+]+)/g;
 export const obsidianMultilineCommentRegex = /^%%\n[^%]*\n%%/gm;
 export const wordSplitterRegex = /[,\s]+/;
 export const ellipsisRegex = /(\. ?){2}\./g;
@@ -100,4 +99,8 @@ export function getFirstHeaderOneText(text: string) {
   }
 
   return '';
+}
+
+export function matchTagRegex(text: string): string[] {
+  return [...text.matchAll(tagWithLeadingWhitespaceRegex)].map((match) => match[2]);
 }
--- a/src/utils/regex.ts
+++ b/src/utils/regex.ts
@@ -14,6 +14,7 @@ export const wikiLinkRegex = /(!?)\[{2}([^\][\n|]+)(\|([^\][\n|]+))?\]{2}/g;
 // based on https://davidwells.io/snippets/regex-match-markdown-links
 export const genericLinkRegex = /(!?)\[([^[]*)\](\(.*\))/g;
 export const tagRegex = /(?<=\s|^)#[^\s#;.,><?!=+]+/g;
+export const tagWithLeadingWhitespaceRegex = /(?:\s|^)#[^\s#;.,><?!=+]+/g;
 export const obsidianMultilineCommentRegex = /^%%\n[^%]*\n%%/gm;
 export const wordSplitterRegex = /[,\s]+/;
 export const ellipsisRegex = /(\. ?){2}\./g;
--- a/src/utils/regex.ts
+++ b/src/utils/regex.ts
@@ -13,7 +13,7 @@ export const codeBlockRegex = new RegExp(`${backtickBlockRegexTemplate}|${tildeB
 export const wikiLinkRegex = /(!?)\[{2}([^\][\n|]+)(\|([^\][\n|]+))?\]{2}/g;
 // based on https://davidwells.io/snippets/regex-match-markdown-links
 export const genericLinkRegex = /(!?)\[([^[]*)\](\(.*\))/g;
-export const tagRegex = /(?:\s|^)#[^\s#;.,><?!=+]+/g;
+export const tagRegex = /(?<=\s|^)#[^\s#;.,><?!=+]+/g;
 export const obsidianMultilineCommentRegex = /^%%\n[^%]*\n%%/gm;
 export const wordSplitterRegex = /[,\s]+/;
 export const ellipsisRegex = /(\. ?){2}\./g;
--- a/__tests__/yaml-title.test.ts
+++ b/__tests__/yaml-title.test.ts
@@ -127,5 +127,19 @@ ruleTest({
         # Title
       `,
     },
+    {
+      testName: 'Tag after first header is not considered as a part of the title',
+      before: dedent`
+        # title
+        #tag
+      `,
+      after: dedent`
+        ---
+        title: title
+        ---
+        # title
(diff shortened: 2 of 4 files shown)


src/utils/regex.ts
```diff
@@ -13,7 +13,8 @@ export const codeBlockRegex = new RegExp(`${backtickBlockRegexTemplate}|${tildeB
 export const wikiLinkRegex = /(!?)\[{2}([^\][\n|]+)(\|([^\][\n|]+))?\]{2}/g;
 // based on https://davidwells.io/snippets/regex-match-markdown-links
 export const genericLinkRegex = /(!?)\[([^[]*)\](\(.*\))/g;
-export const tagRegex = /(?:\s|^)#[^\s#;.,><?!=+]+/g;
+export const tagRegex = /(?<=\s|^)#[^\s#;.,><?!=+]+/g;
```
pjkaufman, line 16: I definitely cannot merge this for 2 reasons:
1. Why do we need to check for tags with leading whitespace?
2. Lookbehinds do not work on iOS, so this is not cross-platform compatible.

pjkaufman (changes requested):

src/utils/regex.ts
pjkaufman: I would recommend fixing the header regex which is the root of the problem: https://github.com/platers/obsidian-linter/blob/master/src/utils/regex.ts#L91

mnaoumov: 1. To fix tests in `Move Tags to Yaml` rule, where tag should be removed together with the leading whitespace
2. I didn't know about iOS regex. I will think how to rewrite without it

pjkaufman: > 1. To fix tests in `Move Tags to Yaml` rule, where tag should be removed together with the leading whitespace
> 2. I didn't know about iOS regex. I will think how to rewrite without it

I must be missing something because the tests currently work for `Move Tags to Yaml` and there was no change to it directly here in the actual logic. So changing how tags are parsed does not seem to me to make sense as the fix to the problem. We could ignore tags again for the `Yaml Tile` and `Yaml Title Alias` rules, but I would much rather have a solution where it only grabs the header like it is supposed to.

mnaoumov: > I would recommend fixing the header regex which is the root of the problem: https://github.com/platers/obsidian-linter/blob/master/src/utils/regex.ts#L91

No, this is not going to help, because by the time we are executing this regex, our header is transformed into

`# title#tag-placeholder`

At this stage we have no idea that this `#tag-placeholder` should be trimmed

mnaoumov: Of course, we can hard-code trimming `#tag-placeholder`, but for me it looks like a leaking abstraction

mnaoumov: > > 1. To fix tests in `Move Tags to Yaml` rule, where tag should be removed together with the leading whitespace
> > 2. I didn't know about iOS regex. I will think how to rewrite without it
> 
> I must be missing something because the tests currently work for `Move Tags to Yaml` and there was no change to it directly here in the actual logic. So changing how tags are parsed does not seem to me to make sense as the fix to the problem. We could ignore tags again for the `Yaml Tile` and `Yaml Title Alias` rules, but I would much rather have a solution where it only grabs the header like it is supposed to.

tests for `Move Tags to Yaml` work because of the fix I made https://github.com/platers/obsidian-linter/pull/552/files#diff-31442bc8965fd62faa01fcec5e0684b8e099aec107795c975208f90eec712256R94

pjkaufman: Ah. I see. Does the following regex work: `/#[^\s#;.,><?!=+]+/g`? I am trying to remember by it needed to either start a file or be preceded by a hashtag (likely due to a multiple hashtag combination).

pjkaufman: The following should work as expected: `/(?:^|[\t\v\f\r \u00a0\u2000-\u200b\u2028-\u2029\u3000])#[^\s#;.,><?!=+]+/m`. It is essentially all whitespace except the newline character.

pjkaufman: It has the added bonus of avoiding the issue of avoiding lookbehinds.

mnaoumov: > Ah. I see. Does the following regex work: `/#[^\s#;.,><?!=+]+/g`? I am trying to remember by it needed to either start a file or be preceded by a hashtag (likely due to a multiple hashtag combination).

I think this regex will be false positive on strings like `a#b`

pjkaufman: > > Ah. I see. Does the following regex work: `/#[^\s#;.,><?!=+]+/g`? I am trying to remember by it needed to either start a file or be preceded by a hashtag (likely due to a multiple hashtag combination).
> 
> I think this regex will be false positive on strings like `a#b`

Gotcha the suggestion after that should not suffer from the same flaws: `/(?:^|[\t\v\f\r \u00a0\u2000-\u200b\u2028-\u2029\u3000])#[^\s#;.,><?!=+]+/m`

mnaoumov: > The following should work as expected: `/(?:^|[\t\v\f\r \u00a0\u2000-\u200b\u2028-\u2029\u3000])#[^\s#;.,><?!=+]+/m`. It is essentially all whitespace except the newline character.

I tried a similar but simpler trick `[^\S\n]` but they won't work either and other tests will break. We need indeed to find the tags including after line break, but not include the whitespace in the capture.

Without lookbehind, it requires additional logic

pjkaufman: The only test that fails locally for me is the following: `Move Tags to Yaml › Make sure that removing tags respects ignore list` which only fails because it does not remove the previous character anymore.

pjkaufman: That is kind of what I would expect since we are expressly saying that the regex used for ignore no longer includes whitespace. We can do something like what you did and swap out the regex used where the old regex is used in `Move Tags to Yaml` and the actual tag replacement in the ignore types uses the new regex using the global flag.

pjkaufman: It would not be a one size fits all use case, but it would do the trick and be clear enough to work from what I can tell. But maybe it would break something else.

pjkaufman: Add this to `ignore-types.ts`

``` Typescript
/**
 * Replaces all obsidian tags in the given text with a placeholder.
 * @param {string} text The text to replace tags in
 * @param {string} tagPlaceholder The placeholder to use for obsidian tags
 * @return {string} The text with tags replaced
 * @return {string[]} The obsidian tags replaced
 */
function replaceTags(text: string, tagPlaceholder: string): IgnoreResults {
  const tagMatches: string[] = [];
  text = text.replaceAll(tagRegex, (tagPlusSpace: string) => {
    const hashtagIndex = tagPlusSpace.indexOf('#');

    tagMatches.push(tagPlusSpace.substring(hashtagIndex));

    return tagPlusSpace.substring(0, hashtagIndex) + tagPlaceholder;
  });

  return {newText: text, replacedValues: tagMatches};
}
```

and then update the `ignore-type` list and as follows and it should fix the issue: `tag: {replaceAction: replaceTags, placeholder: '#tag-placeholder'},`

mnaoumov:
@pjkaufman I tried to come up with a more universal solution, so we explicitly name `tagWithLeadingWhitespaceRegex` and have helper method `matchTagRegex` to extract tags without whitespaces. So far it is not absolutely required, and we could just leave `Move tags to YAML` as is, but I think we might benefit from such helpers in future development

pjkaufman:
This looks fine to me.

pjkaufman (approved):

PR #582, opened 2023-01-12
Fix bugs that might make lint on a correct timestamp tag

### LINKED ISSUE #580: FR: Enable only update create time if create time doesn't exist
Last week, I accidently deleted all my notes, after I restored all my notes, the metadata of create time and update time of my notes are totally changed to the new time, if I lint my notes, the original create_time yaml tag and update_time yaml tag will simultaneously be linted, however, I want to preserve the create_time yaml tag and only lint update time tag since the create time tag is already correct, which is important for me to remember the first time I created the notes.

I wish there was an option that would let me choose whether I would like not to lint create time if create_time tag already exists, otherwise that would lint simultaneouly update create time and update_time tag.

aiyolo:
Fixes #580 
Change Made:
- remove extra spaces in `option.format` which might cause `createdDateTime.isValid()` return false even if `createDateTime` is correct.


pjkaufman:
The main problem I am seeing with this is you are trimming more than just the format. In doing so, I am not sure what exactly you are trying to fix. This sounds more like you need to use a tool beyond the Linter to fix the invalid addition of whitespace as the Linter is not designed to handle extra whitespace in the actual value itself.

The format can be trimmed, but that would need to be done prior to the rule or you will create this problem anywhere the time format is used. It would likely need to be on update and save.

pjkaufman:
I have recently looked at this and I don't believe that this change makes sense for the codebase. It is true that you can trim whitespace after the format if you would like. However I do not believe that we should force users to not be able to do so if they so choose. To fix the issue with the whitespace being present, I recommend using a regex replacement for all of the files in your vault. Sorry if this was not what you were hoping for. I just don't know that this is the right solution for the exact scenario in question.

aiyolo:
Thanks for your reply, I almost understood what you mean. 
Still, I hope that this pull request can be merged. As for the reasons, firstly, whitespaces at the end of the moment format are usually meaningless and trimming the whitespaces has no effect on the generated timestamp; secondly, I don't think users will add whitespaces at the end of the moment format on purpose. For me, I originally want to set the format as "YYYY-MM-DD HH:mm:ss", but I typed an extra whitespace by accident, and the moment format becomes "YYYY-MM-DD HH:mm:ss ". It is really difficult to find the error. The right solution I think to solve the problem perhaps is to add an error hint for some incorrect moment format, for example, if you add extra whitespaces, there may be a red tilde around the input box. 
If you still believe this change makes no sense, it is okay with me.

aiyolo:
We should really try to use the time data in the file front-matter yaml first, if they don't exist then we use the file metadata, because our notes may contain notes exported by other note-taking software, and the file metadata of these exported notes may be lost during copying.

pjkaufman:
> We should really try to use the time data in the file front-matter yaml first, if they don't exist then we use the file metadata, because our notes may contain notes exported by other note-taking software, and the file metadata of these exported notes may be lost during copying.

This comment may be in the wrong place, but if it is not, that functionality was added as an option for users to select. I am just not sure if it has been released yet.

aiyolo:
I don't think it has been released. Currently, we can only update both created time and updated time by tracking metadata of files.  If we have a note exported from another note-taking app that have a created time frontmatter tag different from the metadata information, then the original created time will definitively be lost when we lint on the file. 
So, my suggestion is to preserve the original created time frontmatter as much as possible when we lint on our notes.

pjkaufman:
I am going to close this since I believe that whitespace being added to the end of the timestamp is a valid timestamp format that people can specify if they so choose. However if this does have more people who want this change, we may end up revisiting this in the future.

PR #615, opened 2023-02-02
Perform escaping validation

### LINKED ISSUE #614: Bug: `YAML Title Alias` rule fails if the note name starts with exclamation mark `!`
## Describe the Bug

`YAML Title Alias` rule fails if the note name starts with exclamation mark `!`

## How to Reproduce

1. Create a note named `!Note.md`
2. Apply `YAML Title Alias` rule to it

## Expected Behavior

No errors

## Actual Behavior

```
logger.ts:21 [Obsidian Linter] Lint File Error in File '!Note.md':
Error "YAML Title Alias" encountered an error in the yaml:  unknown tag !<!Note> (2:1)

 1 | title: !Note
 2 | 
-----^
YAMLException: unknown tag !<!Note> (2:1)

 1 | title: !Note
 2 | 
-----^
    at generateError (plugin:obsidian-linter:2695:10)
    at throwError (plugin:obsidian-linter:2698:9)
    at composeNode (plugin:obsidian-linter:3603:7)
    at readBlockMapping (plugin:obsidian-linter:3353:11)
    at composeNode (plugin:obsidian-linter:3546:78)
    at readDocument (plugin:obsidian-linter:3678:3)
    at loadDocuments (plugin:obsidian-linter:3720:5)
    at load$1 (plugin:obsidian-linter:3738:19)
    at loadYAML (plugin:obsidian-linter:4602:23)
    at YamlTitleAlias.apply (plugin:obsidian-linter:53076:24)
```

mnaoumov:
Fixes #614 

I had to add a hack to skip the new logic for `Escape YAML Special Characters` rule, because the new approach would not work fine with abovementioned rule.

The idea was to ensure that in difficult scenarios use `js-yaml` serialization techniques.

So if we have a title `- title` we see that it deserializes as an array instead of string, so we conclude that we need to enclose the title in quotes `"- title"`.

However such logic is fine when we are sure that we were escaping only the string.

But `Escape YAML Special Characters` is executed when the entire YAML object is processed, so each line is not meant to be a string, therefore abovementioned logic would not make sense.

I'd prefer `Escape YAML Special Characters` rule to be refactored to take advantage of the improvements I've made, but I think this would require rewriting that rule entirely, ideally treating the YAML as an AST, not by manual parsing line-by-line.

I also noticed that `Escape YAML Special Characters` might produce an invalid YAML, which in my opinion, we should avoid by all means.

### DIFF
--- a/src/rules/escape-yaml-special-characters.ts
+++ b/src/rules/escape-yaml-special-characters.ts
@@ -1,7 +1,7 @@
 import {Options, RuleType} from '../rules';
 import RuleBuilder, {BooleanOptionBuilder, ExampleBuilder, OptionBuilderBase} from './rule-builder';
 import dedent from 'ts-dedent';
-import {escapeStringIfNecessaryAndPossible, formatYAML, QuoteCharacter as QuoteCharacter} from '../utils/yaml';
+import {escapeStringIfNecessaryAndPossible, formatYAML, QuoteCharacter} from '../utils/yaml';
 
 class EscapeYamlSpecialCharactersOptions implements Options {
   @RuleBuilder.noSettingControl()
--- a/src/rules/escape-yaml-special-characters.ts
+++ b/src/rules/escape-yaml-special-characters.ts
@@ -1,11 +1,11 @@
 import {Options, RuleType} from '../rules';
 import RuleBuilder, {BooleanOptionBuilder, ExampleBuilder, OptionBuilderBase} from './rule-builder';
 import dedent from 'ts-dedent';
-import {escapeStringIfNecessaryAndPossible, formatYAML} from '../utils/yaml';
+import {escapeStringIfNecessaryAndPossible, formatYAML, QuoteCharacter as QuoteCharacter} from '../utils/yaml';
 
 class EscapeYamlSpecialCharactersOptions implements Options {
   @RuleBuilder.noSettingControl()
-    defaultEscapeCharacter?: string = '"';
+    defaultEscapeCharacter?: QuoteCharacter = '"';
   tryToEscapeSingleLineArrays?: boolean = false;
 }
 
@@ -84,7 +84,7 @@ export default class EscapeYamlSpecialCharacters extends RuleBuilder<EscapeYamlS
                 arrayItem = arrayItem.substring(0, arrayItem.length - 1).trimEnd();
               }
 
-              arrayItems[j] = arrayItems[j].replace(arrayItem, escapeStringIfNecessaryAndPossible(arrayItem, options.defaultEscapeCharacter));
+              arrayItems[j] = arrayItems[j].replace(arrayItem, escapeStringIfNecessaryAndPossible(arrayItem, options.defaultEscapeCharacter, false, true));
             }
 
             yamlLines[i] = yamlLines[i].replace(value, '[' + arrayItems.join(',') + ']');
@@ -93,7 +93,7 @@ export default class EscapeYamlSpecialCharacters extends RuleBuilder<EscapeYamlS
           continue;
         }
 
-        yamlLines[i] = yamlLines[i].replace(value, escapeStringIfNecessaryAndPossible(value, options.defaultEscapeCharacter));
+        yamlLines[i] = yamlLines[i].replace(value, escapeStringIfNecessaryAndPossible(value, options.defaultEscapeCharacter, false, true));
       }
 
       return yamlLines.join('\n');
--- a/src/utils/yaml.ts
+++ b/src/utils/yaml.ts
@@ -1,4 +1,4 @@
-import {load} from 'js-yaml';
+import {load, dump} from 'js-yaml';
 import {getTextInLanguage} from '../lang/helpers';
 import {escapeDollarSigns, yamlRegex} from './regex';
 
@@ -94,6 +94,8 @@ export enum NormalArrayFormats {
   MultiLine = 'multi-line',
 }
 
+export type QuoteCharacter = '\'' | '"';
+
 /**
  * Formats the yaml array value passed in with the specified format.
  * @param {string | string[]} value The value(s) that will be used as the parts of the array that is assumed to already be broken down into the appropriate format to be put in the array.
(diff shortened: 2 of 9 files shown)


pjkaufman:
The reason `js-yaml` isn't used is because Obsidian can create values that are invalid in regular JS which would not allow proper Obsidian YAML escaping for tags and aliases.

As for the other approach,  I don't like using a try catch to see if something needs escaping personally.

I agree that escape yaml characters could have used it at 1 point, but it escapes some cales that don't always need escaping in YAML thus it would break functionality to use `js-yaml`. This is just my cursory glance at this though.

pjkaufman:
Also could you explain the following?

> I also noticed that Escape YAML Special Characters might produce an invalid YAML, which in my opinion, we should avoid by all means.

pjkaufman:
If we plan to use `js-yaml`, I would prefer we handle these scenarios for everything rather than just this one case.

For `tag`, `tags`, `alias`, and `aliases` we could probably manually parse them since they should be valid anyways. I am curios to hear your suggestions around this.

pjkaufman:
> > Also could you explain the following?
> > > I also noticed that Escape YAML Special Characters might produce an invalid YAML, which in my opinion, we should avoid by all means.
> 
> The case when the string already has both single and double quotes are not processed by the previous implementation. There is even a test case that claims that it produces an invalid YAML

I am curious which one you are referring to. It does state that if you try to force the escaping of a single-line YAML array this can happen. However that is a problem inherent to the fact that Obsidian adds new kinds of YAML arrays.

I am not opposed to moving to using an AST parsing implementation, but I have yet to see how this fixes problems with invalid YAML and properly recognizing the new array types added for the 4 specific keys (`tag`, `tags`,  `alias`, `aliases`).

pjkaufman:
I have not yet decided on the implementation side of things, but I am opposed to the try catch method of saying if it throws an error it needs escaping since this is bad for performance and there should be another way to fix invalid strings without hitting the exception.

mnaoumov:
> I have not yet decided on the implementation side of things, but I am opposed to the try catch method of saying if it throws an error it needs escaping since this is bad for performance and there should be another way to fix invalid strings without hitting the exception.

As far as I see `js-yaml` doesn't have a `validate` function to check if the input is the valid YAML or not.  That's why I had to deal with the `try-catch` approach.

I doubt the severity of the performance impact. 99% of the cases, the titles are relatively short and without special characters, so it will fall into basic case.

mnaoumov:
@pjkaufman can you please elaborate what do you call
> a new array type

?

pjkaufman:
> @pjkaufman can you please elaborate what do you call
> 
> > a new array type
> 
> ?

A new array type are things that normally are not considered arrays, but Obsidian treats them as arrays and as such in the YAML we would need to treat them that way as well.
The following are all considered arrays in Obsidian, but are not in YAML (unless something changed since I last looked at the specs where only single-line and multi-line arrays existed):
```
tags: tag1 tag2 tag3 tag4 # 4 element array
tags: [tag1 tag2 tag3 tag4] # 4 element array
tags: tag1, tag2, tag3, tag4 # 4 element array
```
I think I am missing one of the special array formats for YAML for Obsidian as well, but these 3 should suffice. We can manually parse them out as they would be considered valid YAML, but it makes using an AST a bit more of a headache for these new array types as they are either parsed as strings or single element arrays.

src/rules/escape-yaml-special-characters.ts
```diff
@@ -1,11 +1,11 @@
 import {Options, RuleType} from '../rules';
 import RuleBuilder, {BooleanOptionBuilder, ExampleBuilder, OptionBuilderBase} from './rule-builder';
 import dedent from 'ts-dedent';
-import {escapeStringIfNecessaryAndPossible, formatYAML} from '../utils/yaml';
+import {escapeStringIfNecessaryAndPossible, formatYAML, QuoteCharacter as QuoteCharacter} from '../utils/yaml';
```
pjkaufman, line 4: Why are we including as here and not in other imports?

src/utils/yaml.ts
```diff
@@ -324,10 +326,45 @@ export function isValueEscapedAlready(value: string): boolean {
...
+
+  const escapeWithOtherCharacter = dump(value, {
+    lineWidth: -1,
+    quotingType: defaultEscapeCharacter == '"' ? '\'' : '"',
+    forceQuotes: forceEscape,
+  }).slice(0, -1);
+
+  if (escapeWithOtherCharacter === value || escapeWithOtherCharacter.length < escapeWithDefaultCharacter.length) {
+    return escapeWithOtherCharacter;
+  }
```
pjkaufman, line 362: Could you explain how these checks determine that you should use `escapeWithOtherCharacter`?

My confusion may come from a lack of understanding around `dump`, but it looks to me like the string could need escaping with a double quote (not a single quote) and this would fail to yield the proper result if the default character was `'`. However it may be that `js-yaml` does some weird escaping where it uses `'` to somehow escape a single quote character.

src/rules/escape-yaml-special-characters.ts
mnaoumov: No reason, it is a leftover of the other refactoring. I'll fix it

src/utils/yaml.ts
mnaoumov: @pjkaufman 

I checked the way `dump` works

You call `result = dump(x, { quotingType: y })`

| x   | y   | result |
|-----|-----|--------|
| `"` | `"` | `"\""` |
| `'` | `"` | `"'"`  |
| `"` | `'` | `'"'`  |
| `'` | `'` | `''''` |

I ran all the results above through YAML validator and confirmed they are all valid YAML strings and represent the value in `x`

The idea in my implementation was to see which of the quotes characters provide shorter output and use it.

For the synthetic examples in the table above, we see that when we escape single quote character `'` (lines 2 and 4), it's better to escape with a double quote character `"` (line 2). When we escape double quote character `"` (lines 1 and 3), it's better to escape with a single quote characer `'` (line 3). That matches the logic in the previous implementation

pjkaufman:
I am fine with having this merged, though we do need to fix the merge conflict first.

mnaoumov:
> I am fine with having this merged, though we do need to fix the merge conflict first.

Fixed the merged conflict

pjkaufman:
Thanks!

pjkaufman (approved):

PR #646, opened 2023-03-11
Fix duplicate settings desc

northword:
There is a duplicate settings item description in the debug settings item.

before:

Snipaste_2023-03-11_17-58-23

after: 

Snipaste_2023-03-11_18-41-54

```ts
    /**
     * Renders markdown string to an HTML element.
     * @param markdown - The markdown source code
     * @param el - The element to append to
     * @param sourcePath - The normalized path of this markdown file, used to resolve relative internal links
     * @param component - A parent component to manage the lifecycle of the rendered child components, if any
     * @public
     */
    static renderMarkdown(markdown: string, el: HTMLElement, sourcePath: string, component: Component): Promise<void>;
```

The `el` element is appended rather than replaced, so the original text needs to be removed before processing.


pjkaufman (approved):
This looks fine to me.

Issue #219, opened 2022-06-08
FR: "Heading blank lines" Add Option to Not Add Empty Line At Start of File Or After YAML


anijatsu:
Would it be possible to have an exception for this, either by default or via option?

Example
---
```markdown
---
alias: ['authoughts']
---
# Automatic thoughts
```

pjkaufman:
It is possible, but I am curious what the rationale is behind this. Do you mind providing some rationale @anijatsu ?

pjkaufman:
The fix for this should now be on master and is slated for the next release. Please let us know if for some reason it is not fixed on master or in the next release.

There is a new option that should be available that should allow you to turn off the new line between the Yaml frontmatter and the header.

PR #656, opened 2023-03-18
fix typo in example of heading-blank-lines

northword:
document: https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#heading-blank-lines
related: https://github.com/platers/obsidian-linter/issues/219

> Example: Empty line before header and after Yaml is removed with `Empty Line Between Yaml and Header=true`
> 
> Before:
> ```
> ---
> key: value
> ---
> # Header
> Paragraph here...
> After:
> ```
> 
> ```
> ---
> key: value
> ---
> # Header
> 
> Paragraph here...
> ```

when `Empty Line Between Yaml and Header=true`, it should be 

```
> ---
> key: value
> ---
>
> # Header
> 
> Paragraph here...
``` 

only when `Empty Line Between Yaml and Header=false`, it is 

```
> ---
> key: value
> ---
> # Header
> 
> Paragraph here...
``` 

In addition, a blank line has been added after the YAML in `before` to better illustrate that the blank line between the YAML and the title will be removed when `Empty Line Between Yaml and Header=false`.

### DIFF
--- a/src/rules/heading-blank-lines.ts
+++ b/src/rules/heading-blank-lines.ts
@@ -93,11 +93,12 @@ export default class HeadingBlankLines extends RuleBuilder<HeadingBlankLinesOpti
       }),
       new ExampleBuilder({
         // accounts for https://github.com/platers/obsidian-linter/issues/219
-        description: 'Empty line before header and after Yaml is removed with `Empty Line Between Yaml and Header=true`',
+        description: 'Empty line before header and after Yaml is removed with `Empty Line Between Yaml and Header=false`',
         before: dedent`
           ---
           key: value
           ---
+          
           # Header
           Paragraph here...
         `,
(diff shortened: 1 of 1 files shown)


src/rules/heading-blank-lines.ts
```diff
@@ -93,11 +93,12 @@ export default class HeadingBlankLines extends RuleBuilder<HeadingBlankLinesOpti
       }),
       new ExampleBuilder({
         // accounts for https://github.com/platers/obsidian-linter/issues/219
-        description: 'Empty line before header and after Yaml is removed with `Empty Line Between Yaml and Header=true`',
+        description: 'Empty line before header and after Yaml is removed with `Empty Line Between Yaml and Header=false`',
         before: dedent`
           ---
           key: value
           ---
+
```
pjkaufman, line 101: In order to make this addition you should be setting the blank line like so (I see it was not there before, but that was standard decided upon to help make it clear what whitespace is intended since dedent and eslint rules around whitespace are in use):
```suggestion
          ${''}
```

pjkaufman (changes requested):
If the indentation is fixed as requested by the CI, I am fine with approving this.

northword:
Thanks, the question has been revised, please review again.

pjkaufman (approved):

Issue #560, opened 2022-12-20
Bug: `move-math-block-indicators-to-their-own-line` Incorrectly does not recognize `>` in block quotes as own line


mzietlow:
## Describe the Bug

When `move-math-block-indicators-to-their-own-line` is enabled, Linter moves `$$math-blocks$$` outside of block quote environments.

## How to Reproduce

1. Enable `move-math-block-indicators-to-their-own-line`
2. Create a new Document
3. Write a block quote with math-block, e.g. 
``` markdown
> $$math$$

```
4. Lint the file multiple times. After three steps, the process converges to
``` markdown
> 
$$
math
$$

```

### First Linting Step
``` markdown
> $$
math
$$

```

### Second Linting Step
``` markdown
> 
$$

math

$$

```
### Final Linting Step
``` markdown
> 
$$
math
$$

```

The process has now converged, further linting does not change the file.

## Expected Behavior

The minimal example

``` markdown
> $$math$$

```

Should be converted into

``` markdown
> $$
> math
> $$

```

## Device

I only tested on Desktop

- [x] Desktop
- [ ] Mobile

## Additional Context

The issue does not occur with `empty-line-around-math-blocks`.

pjkaufman:
Thanks for reporting this @mzietlow . I did not account for this when the rule was written. I am not sure I have time to account for it before I do the next release and then take a break until after the end of the year, so this may stick around for a couple of weeks, but hopefully it won't be too hard to fix when I get back.

If anyone wants to tackle this while I am away, I am more than happy to review a PR for a fix to this.

mzietlow:
Don't worry, I am just happy if it gets fixed eventually :) 

Unfortunately I am ill currently, so I cannot set up the dev environment and tackle it myself either. Maybe I can give it a shot earlier next week

pjkaufman:
Gotcha. Hopefully you feel better soon. I was going to take a stab at this yesterday and then I ran into the problem of how to handle the following scenario:

``` markdown
> blockquote text
> > Nested blockquote $$ math $$
> > More text
> Some more text
```

My first impression tells me it should become the following:

``` markdown
> blockquote text
> > Nested blockquote 
> > $$
> >  math 
> > $$
> > More text
> Some more text
```

The only problem would be capturing the beginning of the line. It could be something where it grabs the start of the line by iterating character by character only including whitespace and the blockquote indicator.

I might have time to take a stab at this today, but we will see.

pjkaufman:
[...]

PR #658, opened 2023-03-18
Refactor Logic Around Blank Line Being Put Around Content

pjkaufman:
This makes the empty line around content logic more robust and reusable.

Changes Made:
- Converted the logic over to get the start of the line ignoring non-whitespace and greater than characters (>)
- Made sure the logic would decide which empty line value to use when blockquotes were involved to make things more uniform
- Updated several examples to show the change in logic


pjkaufman:
This was started on because it seemed related to https://github.com/platers/obsidian-linter/issues/560 at the time. Now it is just more of an enhancement.

PR #676, opened 2023-04-08
Fix Issue with Footnote Reorder When There are More Footnote References than Footnotes and the Last Footnote Reference for One of the Footnotes Comes After a Different Footnote Reference

### LINKED ISSUE #626: Bug: move-footnotes-to-the-bottom Incorrectly Changes Order of Footnotes
## Describe the Bug

move-footnotes-to-the-bottom incorrectly changes the order of footnotes at the bottom when multiple footnote references to the same footnote are used.

## How to Reproduce

1. Enable the rule `move-footnotes-to-the-bottom`, and disable the rule `re-index-footnotes`.
2. Create a file with the following text:
```markdown
First,[^1] followed by the Second,[^2] and then the First[^1] again.

[^1]: First Reference
[^2]: Second reference
```
3. Run Linter.
4. The text above is changed to the following:
```markdown
First,[^1] followed by the Second,[^2] and then the First[^1] again

[^2]: Second reference
[^1]: First Reference
```

## Expected Behavior

Don't re-order the footnotes, leave them in order as 1, 2, ...

## Device

- [x] Desktop
- [ ] Mobile

pjkaufman:
Fixes #626 

Changes Made:
- Changed the logic from always just using the last currently unused reference to an index key to determine the order of the footnotes at the end of the file to instead use the earliest unused reference for each footnote using the same key
- Added a test for the scenario in question

### DIFF
--- a/__tests__/move-footnotes-to-the-bottom.test.ts
+++ b/__tests__/move-footnotes-to-the-bottom.test.ts
@@ -173,6 +173,20 @@ ruleTest({
         [^1]:[b](http://b.com)
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/626
+      testName: 'Make sure that footnote references are not accidentally reordered when there are multiple references to the same footnote where the first and last reference to the footnote have other footnote references in between them',
+      before: dedent`
+        First,[^1] followed by the Second,[^2] and then the First[^1] again.
+        ${''}
+        [^1]: First Reference
+        [^2]: Second reference
+      `,
+      after: dedent`
+        First,[^1] followed by the Second,[^2] and then the First[^1] again.
+        ${''}
+        [^1]: First Reference
+        [^2]: Second reference
+      `,
+    },
   ],
 });
-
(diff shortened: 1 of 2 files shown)


redactedscribe:
Does this fix footnotes...

```markdown
Foo[^1]

[^1]: Bar
[^2]: Baz
```

...being re-index into:

```markdown
Foo[^2]

[^1]: Baz
[^2]: Bar
```

?

pjkaufman:
I think it just throws an error now in this scenario, but I am not certain on that @redactedscribe .

redactedscribe:
Just to explain what I expect in the scenario I posted: After re-indexing, Foo should remain `Foo[^1]` and with it Bar and Baz should not be reversed.

pjkaufman:
> Just to explain what I expect in the scenario I posted: After re-indexing, Foo should remain `Foo[^1]` and with it Bar and Baz should not be reversed.

I am seeing 
``` markdown
Foo[^1]

[^1]: Bar
[^2]: Baz
```
before and after linting @redactedscribe . So it should be remaining the same.

PR #688, opened 2023-04-17
Translation settings fields to chinese

windily-cloud:
I have encountered several issues while using the commands you provided:

1. After running `npm ci`, I need to execute `npm run` build to generate the translation-helper file.
2. However, this doesn't generate the style.css file.
3. When I run `npm run translate` and select zh-cn, the corresponding translation doesn't get updated in the corresponding zh-cn file.

As a result, I opted to directly translate the en.ts file. I encountered some changes during the debugging process, so I'd appreciate it if you could help me address those. I'm sorry for the delay in translation as I've been very busy recently.

Lastly, having a developer like you for Obsidian is truly an honor. I  also feel lucky to be able to use linter plugin.


pjkaufman:
> I have encountered several issues while using the commands you provided:
> 
> 1. After running `npm ci`, I need to execute `npm run` build to generate the translation-helper file.
> 2. However, this doesn't generate the style.css file.
> 3. When I run `npm run translate` and select zh-cn, the corresponding translation doesn't get updated in the corresponding zh-cn file.
> 
> As a result, I opted to directly translate the en.ts file. I encountered some changes during the debugging process, so I'd appreciate it if you could help me address those. I'm sorry for the delay in translation as I've been very busy recently.
> 
> Lastly, having a developer like you for Obsidian is truly an honor. I also feel lucky to be able to use linter plugin.

Thanks for letting me know. I will have to check what is up with `npm run translate` on `zh-cn`. I know it worked for Spanish, but maybe that is because it is not a compound name.

pjkaufman:
> > I have encountered several issues while using the commands you provided:
> > 
> > 1. After running `npm ci`, I need to execute `npm run` build to generate the translation-helper file.
> > 2. However, this doesn't generate the style.css file.
> > 3. When I run `npm run translate` and select zh-cn, the corresponding translation doesn't get updated in the corresponding zh-cn file.
> > 
> > As a result, I opted to directly translate the en.ts file. I encountered some changes during the debugging process, so I'd appreciate it if you could help me address those. I'm sorry for the delay in translation as I've been very busy recently.
> > Lastly, having a developer like you for Obsidian is truly an honor. I also feel lucky to be able to use linter plugin.
> 
> Thanks for letting me know. I will have to check what is up with `npm run translate` on `zh-cn`. I know it worked for Spanish, but maybe that is because it is not a compound name.

The issue here seems to have been that it was in the map differently from the name of the file because the file language was listed as `zh`, but it referred to `zh-cn.ts` which would cause the translation helper to be unable to properly save the language values. Sorry about that.

I am going to try to remove any other scenarios like that if possible.

src/lang/helpers.ts
```diff
@@ -53,7 +53,7 @@ export const localeMap: { [k: string]: Partial<LanguageStrings> } = {
   tr,
   uk,
   'zh-TW': zhTW,
-  'zh': zhCN,
+  'zh-cn': zhCN,
```
pjkaufman, line 56: Updating this may fix the translation helper logic, but I think it breaks the language support for Obsidian unless it is just for when you have `zh-cn` selected in the Obsidian dropdown for the language value.

src/lang/locale/zh-cn.ts
```diff
@@ -1,3 +0,0 @@
-// 简体中文
-
-export default {};
```
pjkaufman, line 3: Is there a reason we replaced this file with `zh.ts`?

src/lang/helpers.ts
windily-cloud: The changes I made elsewhere are just to make it work, and to see the effect in obsidian, you can ignore them and just use the files I translated

pjkaufman: Sounds good. I will test this out later on today or tomorrow based on when I get some time.

pjkaufman:
This looks good to me. I will take a look at the actual displaying of the values when I get a chance. Once that happens I will go ahead and merge this.

pjkaufman:
Thanks for taking the time to translate this!

windily-cloud:
> Thanks for taking the time to translate this!

Compared to your efforts, this work of mine is really very insignificant. I've already posted articles in China promoting linter and they all think it works very well. Before linter had i18n, there were already translations by directly modifying the main.js file.

Thanks again for your efforts ！！！！

src/lang/locale/zh-cn.ts
```diff
@@ -1,3 +1,712 @@
...
+  // lint-confirmation-modal.ts
+  'warning-text': '警告',
+  'file-backup-text': '请确保你已备份文件',
+
+  'tabs': {
+    'names': {
+      // tab.ts
+      'general': '通用设置',
+      'custom': '自定义设置',
+      'yaml': 'YAML设置',
```
pjkaufman, line 106: Is YAML supposed to be here?

src/lang/locale/zh-cn.ts
```diff
@@ -1,3 +1,712 @@
...
+      'heading': '标题设置',
+      'content': '脚注设置',
+      'footnote': '内容设置',
+      'spacing': '间距设置',
+      'paste': '粘贴设置',
+      'debug': 'Debug设置',
```
pjkaufman, line 112: Is debug supposed to be here?

pjkaufman (commented):
I tested it out and it works. I noticed one bug in the settings page around language text, so I am putting in a fix for that, but I wanted to check that YAML and Debug are supposed to be in the heading names for the tabs. If so, I will merge this translation in.

pjkaufman:
It works with the following language selected: `简体中文` which maps to `zh` in terms of its language codes.

pjkaufman:
I have yet to hear back on the 2 headings, but I will go ahead and merge this PR and a subsequent PR can be created if either of the 2 headers should not have the English name present.

PR #721, opened 2023-05-14
Cleaned Up List Item Text Update Logic

pjkaufman:
Changes Made:
- Fixed an example that actually included codeblocks to no longer include them
- Refactored and reworked update list item text logic to let unist do the work of finding the list items and then getting the paragraph kids of them as the actual text for them instead of all text inside a list item


pjkaufman:
I need to check to see why the text of the task itself is including the task list indicator. It should not be included, which implies that I need to add the column and offset together to get the position, but I am not certain of this yet. I will test somethings out and see what happens.

PR #725, opened 2023-05-17
Add Custom Ignore Syntax Similar to Prettier's Syntax

### LINKED ISSUE #334: FR: Recognize special blocks or syntaxes to disable Linter on them
I'm starting using [Obsidian_to_Anki plugin](https://github.com/Pseudonium/Obsidian_to_Anki) and it works with an special use of the blank lines and probably other syntaxis. If I use Linter I break these rules and the [Obsidian_to_Anki plugin](https://github.com/Pseudonium/Obsidian_to_Anki) function.

I would like that Linter recognize these rules in the notes and avoid modifying these features of those texts.

Example

Original text
``` markdown
Question %% #flashcard %%
Answer
```
After the current Linter version 
``` markdown
Question %% #flashcard %%

Answer
```
The ideal solution should have a toggle in the configuration to select (or make it default) if we want to respect the rules of [Obsidian_to_Anki plugin](https://github.com/Pseudonium/Obsidian_to_Anki) on these lines (maybe with the toggle to force Linter to act). Another not so good option could accept some regex templates and which rules should not be executed on them.

### LINKED ISSUE #701: FR: add linter-ignore-start,linter-ignore-end

I have some special css fragments, syntax，I have some special css fragments, syntax.

I don't want this area to be formatted

``` markdown
<!-- linter-ignore-start -->
                          This area will not be formatted
<!-- linter-ignore-end -->
```

``` markdown
<!-- linter-ignore-start -->
                          This area will not be formatted
<!-- linter-ignore-end -->```

A clear and concise description of any alternative solutions or features you've considered.

## Additional Context

Add any other context or screenshots about the feature request here.

you can see https://prettier.io/docs/en/ignore.html#range-ignore

### PR DESCRIPTION
**Pull request by pjkaufman**
Fixes #701 
Fixes #334 

Adds an exact syntax for turning off the Linter for certain parts of the document that you do not want linted.

Changes Made:
- Added a new property to the initialization of a rule for options to ignore for the rule which should NOT include custom ignore as that is added under the hood by default
- Custom ignore does not affect paste type rules
- Added eslint ignore pattern for the translation helper that is generated by build
- Added tests for finding custom ignore sections in text
- Updated existing rules to use the new rule ignore type property where possible
- Moved local functions to their own methods in the rules where it seemed easy and feasible to help out with performance as well
- Added some comments to rules with special rule running set to true to give some context as to why they get to run at a special time
- Moved regex to `regex.ts` where it seemed feasible and good to do so
- Swapped `defaultEscapeCharacter` from a string to `QuoteCharacter` where possible since it has a proper type to use now

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 node_modules/
 docs.js
 main.js
+translation-helper.js
 .eslintrc.js
 babel.config.js
--- a/__tests__/disabled-rules.test.ts
+++ b/__tests__/disabled-rules.test.ts
@@ -1,66 +1,88 @@
 import dedent from 'ts-dedent';
 import {getDisabledRules, rules} from '../src/rules';
 
-describe('Disabled rules parsing', () => {
-  it('No YAML', () => {
-    const text = dedent`
+type disabledRulesTestCase = {
+  name: string,
+  text: string,
+  expectedDisabledRules: string[],
+};
+
+const disabledRulesTestCases: disabledRulesTestCase[] = [
+  {
+    name: 'when there is no YAML, no rules are disabled',
+    text: dedent`
       Text
-    `;
-    expect(getDisabledRules(text)).toEqual([]);
-  });
-  it('No ignored rules', () => {
-    const text = dedent`
+    `,
+    expectedDisabledRules: [],
+  },
+  {
+    name: 'when there is no YAML key for disabled rules, no rules are disabled',
+    text: dedent`
       ---
-      ---
-      Text
-    `;
-    expect(getDisabledRules(text)).toEqual([]);
-  });
-  it('Ignore one rule', () => {
-    const text = dedent`
-      ---
-      disabled rules: [ yaml-timestamp ]
-      ---
-      Text
-    `;
-    expect(getDisabledRules(text)).toEqual(['yaml-timestamp']);
-  });
-  it('Ignore some rules', () => {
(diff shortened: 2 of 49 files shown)

### CODE REVIEW

**Comment by pjkaufman** [comment]
I still need to add documentation to the README template and update the template doc and maybe contributing docs to include info on the new syntax.

Testing against my own vault to make sure things did not break is also necessary.

**Comment by pjkaufman** [comment]
I need to take a closer look at the rules relating to list items as it seems that it is adding random spaces in checklist items:
```
- [x] Task 1


- [x] Task 2
- [x] Task 3
```

However it does not seem to happen in all files. It just seems to happen in a few of them.

pjkaufman:
It seems to relate to the new rule around removing whitespace before and or after characters.

pjkaufman:
This seems to be a preexisting bug that I need to deal with. It happens when the contents of the list are to be ignored. It seems in those cases it does not properly recognize the start of the list item. I need to identify the root cause of this.

pjkaufman:
I believe I have a fix for the issue. It should be ready shortly. I will push it up once that is the case.

pjkaufman:
I believe that I have custom ignore for all rules ready. It will be separate from rule specific ignores. I need to just double check that it works and then I will merge this.

pjkaufman:
Looks to be working now. I will go ahead and merge this and create the other ticket for custom ignore ranges for specific fields.

PR #746, opened 2023-06-01
Fix translation in zh-cn.ts

lifeek23:
lint-file.name should be '格式化当前文件', and lint-file.error-message should be '格式化当前文件时出错'


pjkaufman:
Based on what I am seeing from Google Translate, this change looks fine. I have no reason to reject this or request changes so I will merge it.

Thanks @lifeek23 for suggesting the change!

pjkaufman:
The change should go out in the next release.

PR #755, opened 2023-06-10
fix: add blank line between table and heading

### LINKED ISSUE #751: Bug: Empty Line Around Tables not respected
## Describe the Bug

"Empty Line Around Tables" option is enabled in Linter  but when the command is ran it removes all empty lines between a Header and the table

## How to Reproduce

1. Enable "Empty Line Around Tables"
2. In a document (not beginning or end) create a H2, create an empty line and add a table after the header
3. Run Linter
4. Empty line between H2 and table is removed and table is not rendered properly

``` markdown
## Logical operators

| Operator | Meaning                  | Example               |
|----------|--------------------------|-----------------------|
|`==`| Is Equal To              | 3 == 5 gives us False |
|`!=`| Not Equal To             | 3 != 5 gives us True  |
|`>`| Greater Than             | 3 > 5 gives us False  |
|`<` | Less Than                | 3 < 5 gives us True   |
|`>=`| Greater Than or Equal To | 3 >= 5 give us False  |
|`<=`|Less Than or Equal To|3 <= 5 gives us True |
```

#### Log
Running linter
Running Format Tags in YAML
Running Format Yaml Array
Running Move Tags to Yaml
Running Remove Trailing Punctuation in Heading
Running Convert Bullet List Markers
Running Remove Consecutive List Markers
Running Remove Empty List Markers
Running Remove Hyphenated Line Breaks
Running Remove Multiple Spaces
Running Two Spaces Between Lines with Content
Running Compact YAML
Running Consecutive blank lines
Running Empty Line Around Blockquotes
Running Empty Line Around Code Fences
Running Empty Line Around Math Blocks
Running Empty Line Around Tables
Running Heading blank lines
Running Remove Empty Lines Between List Markers and Checklists
Running Remove link spacing
Running Space after list markers
Running Trailing spaces
Running Custom Regex
Running Capitalize Headings
Running Custom Lint Commands

## Expected Behavior
There should be one empty line between the H2 and the table

## Screenshots

Before
image

After
image

## Device

- [ x] Desktop
- [ ] Mobile

qazxcdswe123:
should fix #751

### DIFF
--- a/src/rules/heading-blank-lines.ts
+++ b/src/rules/heading-blank-lines.ts
@@ -35,6 +35,9 @@ export default class HeadingBlankLines extends RuleBuilder<HeadingBlankLinesOpti
     text = text.replace(/^\n+(#+\s.*)/, '$1'); // remove blank lines before first heading
     text = text.replace(/(#+\s.*)\n+$/, '$1'); // remove blank lines after last heading
 
+    // blank line between heading and table, see https://github.com/platers/obsidian-linter/issues/751
+    text = text.replace(/(^#+\s.*)(\n\|)/gm, '$1\n$2');
+
     if (!options.emptyLineAfterYaml) {
       text = text.replace(new RegExp('(' + yamlRegex.source + ')\\n+(#+\\s.*)'), '$1\n$5');
     }
--- a/__tests__/heading-blank-lines.test.ts
+++ b/__tests__/heading-blank-lines.test.ts
@@ -103,5 +103,39 @@ ruleTest({
         blah blah blah
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/751
+      testName: 'Blank line between heading and table',
+      before: dedent`
+        # H1
+        | a | b |
+
+        # H1
+
+        | a | b |
+
+
+        # H1
+
+
+        | a | b |
+      `,
+      after: dedent`
+        # H1
+
+        | a | b |
+
+        # H1
+
+        | a | b |
+
+        # H1
+
+        | a | b |
+      `,
+      options: {
+        bottom: false, // this bug only affects bottom false
+        emptyLineAfterYaml: true,
+      },
+    },
   ],
 });


pjkaufman:
Thanks for putting in a fix for this. I need to double check what is going on here. Based on the comments that are present in this PR it looks like we may want to take a different route to fix this (i.e. have the blank lines around tables run after the heading rule). I need to double check things to make sure that is the case.

The blank lines around headings should not have to factor in the type of content that follows it. If it says there is no blank line below it, it should remove the blank line.

src/rules/heading-blank-lines.ts
```diff
@@ -35,6 +35,9 @@ export default class HeadingBlankLines extends RuleBuilder<HeadingBlankLinesOpti
     text = text.replace(/^\n+(#+\s.*)/, '$1'); // remove blank lines before first heading
     text = text.replace(/(#+\s.*)\n+$/, '$1'); // remove blank lines after last heading
 
+    // blank line between heading and table, see https://github.com/platers/obsidian-linter/issues/751
+    text = text.replace(/(^#+\s.*)(\n\|)/gm, '$1\n$2');
+
```
pjkaufman, line 40: Is there a reason to add this specifically to heading blank lines? I am not following why we are trying to account for tables here if the rule says to remove blank lines after headings. That is expressly being requested? This will affect scenarios where blank lines around tables is not enabled.

qazxcdswe123: Ohhhh I see. I just found that there is a [empty line around table rule](https://github.com/platers/obsidian-linter/blob/master/docs/rules.md#empty-line-around-tables). Unfortunately, according to the debug trace, it runs **before** this `heading blank lines` rule. After it adds empty line around table, it gets removed because of this rule.

Probably the `heading-blank-lines` rule should be prioritized? This can simply fix this issue. But for others who don't know there is such a rule like `empty line around table`, maybe add a option (or reminder) to enable `empty line around table`rule?

pjkaufman:
I am closing this in favor of https://github.com/platers/obsidian-linter/pull/760.

PR #756, opened 2023-06-10
fix npm dependencies issues

qazxcdswe123:
unist-util-visit is `src/utils/mdast.ts` and @jest/types is need for jest

I'm not sure this is intentional or by accident, but I need this two dependencies to work. 

Not sure if this is needed. Feel free to close it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "ts-node": "^10.9.1",
     "tslib": "^2.4.1",
     "typescript": "^5.0.4",
-    "unified-lint-rule": "^2.1.1"
+    "unified-lint-rule": "^2.1.1",
+    "@jest/types": "^29.5.0"
   },
   "dependencies": {
     "@popperjs/core": "^2.11.6",
@@ -63,6 +64,7 @@
     "micromark-util-combine-extensions": "^1.0.0",
     "moment-parseformat": "^4.0.0",
     "quick-lru": "^6.1.1",
-    "ts-dedent": "^2.2.0"
+    "ts-dedent": "^2.2.0",
+    "unist-util-visit": "^4.1.2"
   }
 }


package.json
```diff
@@ -63,6 +64,7 @@
     "micromark-util-combine-extensions": "^1.0.0",
     "moment-parseformat": "^4.0.0",
     "quick-lru": "^6.1.1",
-    "ts-dedent": "^2.2.0"
+    "ts-dedent": "^2.2.0",
+    "unist-util-visit": "^4.1.2"
```
pjkaufman, line 68: This should already be included by another package. I say that because it has run just fine up until now. I do not mind adding it, but I am not sure why we would need to add it now.

package.json
```diff
@@ -45,7 +45,8 @@
     "ts-node": "^10.9.1",
     "tslib": "^2.4.1",
     "typescript": "^5.0.4",
-    "unified-lint-rule": "^2.1.1"
+    "unified-lint-rule": "^2.1.1",
+    "@jest/types": "^29.5.0"
```
pjkaufman, line 49: This addition looks perfectly fine to me.

package.json
pjkaufman: It could be that these changes were somehow included globally for me and that is why they work for me. But they seem to work in `main.js` for the plugin. Could you let us know what editor you used @qazxcdswe123 that needed these npm packages to work?

Typically I use VSCode or NeoVim. So it could be that one of them already accounted for the discrepancy in dependencies.

qazxcdswe123: This is what I get after removing this two dependencies, the build/test can't be run. 

I use neovim and the lsp also compliant about the import.

image

image

pjkaufman: Gotcha. It should be fine to merge this then. My node modules locally must have kept those values somehow.

qazxcdswe123: BTW, mayble a different version of this dependency is needed since the UI somehow broke for me 😢

PR #761, opened 2023-06-16
Fix Retention of Created Date Causing `YAML Timestamp` to Always Update Date Modified and Fix Logging of Timing Info

### LINKED ISSUE #745: Bug: File Menu Lint Updates Date Modified in All Cases
Linting a file updates the yaml modification date even though the file hasn't been modified.

**For example:**

File x has modified date in yaml: 2023-04-07T13:56:26+02:00, and on file system: 2023-04-27 17:44.

After linting, the yaml modified date is set to the current date time, rather than the actual last modified date, which was recorded by the file system as 2023-04-27 17:44. 

So in this case I expected the yaml to be updated to the last modification date from the file system rather than the current time.

**Example 2:**

The yaml modified date and file system are the same — a date in the past. In this case the file shouldn't be touched. 

---

This issue occured to me as I was just modifying a few files in a folder of many files and I linted the whole folder, expecting that only the files I had modified would be linted, but every single file got their yaml modification dates updated. So this means that all mod dates for these files are now incorrect.

pjkaufman:
Fixes #745 

Changes Made:
- Added UT to reproduce the scenario in question
- Updated the logic for handling date created updates to make sure that it only updates the date created if the value is missing or is in the wrong format
  - Previously it assumed that if you had the retention of the current value set, it always needed to do an update
  - This is actually incorrect, so I updated the logic to only update the value if the format changed
- Updated time logging logic to reflect the fact that it should only log timing info if debug or lower is set

### DIFF
--- a/__tests__/yaml-timestamp.test.ts
+++ b/__tests__/yaml-timestamp.test.ts
@@ -332,5 +332,44 @@ ruleTest({
         alreadyModified: false,
       },
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/745
+      testName: 'When no changes are made, and force retention of creation date is active, do not update date modified when no change in modification time has been made',
+      before: dedent`
+        ---
+        tag: tag1
+        modified: Tuesday, February 4th 2020, 6:00:00 pm
+        created: Wednesday, January 1st 2020, 12:00:00 am
+        location: "path"
+        ---
+      `,
+      after: dedent`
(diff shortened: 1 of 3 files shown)


aiyolo:
I believe my previous issue https://github.com/platers/obsidian-linter/pull/582 is solved in this PR.

pjkaufman:
That is good to hear.

PR #779, opened 2023-06-29
Label for Custom Replacement

### LINKED ISSUE #774: FR: Custom Rules should get a label

When you have quite a few custom rules with long regexes, it can get quite confusing quickly what exactly those rules do
CleanShot 2023-06-26 at 22 14 25@2x

a field to add a label or note what the rule does

Keeping an extra note in my vault explaining what my rules do 🙈

pjkaufman:
Fixes #774 

Changes Made:
- Added several classes around how things should be styled into 2 rows
- Added a label for custom replacement and made it about 50% of the width of a row


pjkaufman:
I need to add the label value in all the current languages via Google Translate and we should be good to merge this after that.

PR #788, opened 2023-07-11
Ignore Footnote Definitions for Paragraph Blank Lines

### LINKED ISSUE #787: Bug: Paragraph blank lines incorrectly adding blank lines between footnotes
- [ ] I have verified that I am on the latest version of the Linter

## Describe the Bug

When adding footnotes to the bottom of a note, they are normally added without blank lines in between them, e.g.:

image

When saving the file, Linter will then add a bunch of blank lines between the footnotes:

image

This is not required, and looks less appealing than footnotes without the line spaces. Footnotes should be excluded from this linter action.

## How to Reproduce

1. Add 2 or more footnotes to a note, without blank lines between them
2. Save the note, or trigger linter
3. Observe the blank lines added in between each footnote

Or

``` markdown
example here
```

Please make sure to include logs and the rule(s) that cause the issue if applicable. You can include logs by going to the debug tab of the Linter and setting the log level to debug. Then enable the capturing of logs and either lint on save or lint the current file. Then go back to the debug tab and you can grab Linter logs from the last run.

If you are not sure which rule(s) may cause the issue, please attach the `data.json` file located at `<vault>/.obsidian/plugins/obsidian-linter/` or copy the config from the debug tab.

## Expected Behavior

Footnotes should be excluded from this type of linting, or an option should be adding so the user can choose their preference.

Expected output if applicable:

``` markdown
expected output here
```

## Device

- [x] Desktop
- [ ] Mobile

pjkaufman:
Fixes #787 

Changes Made:
- Added a test for the scenario in question
- Refactored the logic around the regex for a footnote definition indicator at the start of a line
- Checked that the beginning of the first line of a paragraph is not a footnote definition

### DIFF
--- a/__tests__/paragraph-blank-lines.test.ts
+++ b/__tests__/paragraph-blank-lines.test.ts
@@ -328,5 +328,26 @@ ruleTest({
         ## Agenda
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/787
+      testName: 'Make sure that an empty list indicator does not have an extra empty line added around it',
+      before: dedent`
+        - reference to footnote 1 [^1]
+        - reference to footnote 2 [^2]
+        - reference to footnote 3 [^3]
+        ${''}
+        [^1]: [github.com](https://github.com)
+        [^2]: [github.com](https://github.com)
+        [^3]: [github.com](https://github.com)
+      `,
+      after: dedent`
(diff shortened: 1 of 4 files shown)


pjlewisuk:
Thanks for the speedy fix, I'll look out for the next release :)

PR #793, opened 2023-07-13
update Chinese translate

Ooopz:
A better translate than Google, fix some error and enhance text consistency.

### DIFF
--- a/__tests__/yaml-key-sort.test.ts
+++ b/__tests__/yaml-key-sort.test.ts
@@ -66,5 +66,40 @@ ruleTest({
         yamlSortOrderForOtherKeys: 'Ascending Alphabetical',
       },
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/796
+      testName: 'Sort works when priority keys end in a colon',
+      before: dedent`
+        ---
+        language: Typescript
+        type: programming
+        tags: computer
+        keywords: []
+        status: WIP
+        date: 01/15/2022
+        modified: Wednesday, January 1st 2020, 12:00:00 am
+        ---
+      `,
+      after: dedent`
+        ---
+        date: 01/15/2022
+        type: programming
+        language: Typescript
+        tags: computer
+        keywords: []
+        status: WIP
+        modified: Thursday, January 2nd 2020, 12:00:00 am
+        ---
+      `,
+      options: {
+        yamlKeyPrioritySortOrder: [
+          'date:',
+          'type:',
+          'language:',
+        ],
+        dateModifiedKey: 'modified',
+        currentTimeFormatted: 'Thursday, January 2nd 2020, 12:00:00 am',
+        yamlTimestampDateModifiedEnabled: true,
+      },
+    },
   ],
 });
--- a/src/lang/locale/en.ts
+++ b/src/lang/locale/en.ts
@@ -304,7 +304,7 @@ export default {
     // empty-line-around-blockquotes.ts
     'empty-line-around-blockquotes': {
       'name': 'Empty Line Around Blockquotes',
-      'description': 'Ensures that there is an empty line around blockquotes unless they start or end a document. **Note that an empty line is either one less level of nesting for blockquotes or a newline character.**',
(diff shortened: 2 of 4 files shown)


Ooopz:
I have test the translate, and it looks good. By the way I think menu structure need an update.

pjkaufman:
> I have test the translate, and it looks good. By the way I think menu structure need an update.

Could you elaborate on what you mean by a menu update being needed?

Ooopz:
In content setting, the rules may need reordered.
better order as below:
Auto-correct Common Misspellings,
Blockquote Style,
Emphasis Style,
Strong Style,
Quote Style,
Ordered List Style,
Unordered List Style,
Convert Bullet List Markers,
Remove Consecutive List Markers,
Remove Empty List Markers,
Remove Hyphenated Line Breaks,
No Bare URLs, 
Proper Ellipsis

Remove Multiple Spaces and Two Spaces Between Lines with Content can move to spacing setting

pjkaufman:
> In content setting, the rules may need reordered. better order as below: Auto-correct Common Misspellings, Blockquote Style, Emphasis Style, Strong Style, Quote Style, Ordered List Style, Unordered List Style, Convert Bullet List Markers, Remove Consecutive List Markers, Remove Empty List Markers, Remove Hyphenated Line Breaks, No Bare URLs, Proper Ellipsis
> 
> Remove Multiple Spaces and Two Spaces Between Lines with Content can move to spacing setting

Generally speaking the order is alphabetical. However I do agree with the comments about the spacing values.

Ooopz:
I haven't notice alphabetical order because I'm using Chinese setting... XD

pjkaufman:
> I haven't notice alphabetical order because I'm using Chinese setting... XD

Gotcha. It may not be visible in other languages.

Ooopz:
conflict has been solved, it's ready for merge.

pjkaufman:
I will take a look. It looks like there are a decent bit of changes, so I need to see about comparing what the meanings are before and after. I don't know Chinese, so the best I can do for validation is check with Google Translate to see how it turns out.

pjkaufman (changes requested):
Thanks for putting in the work to help better the Chinese translation!

I notice a couple of things that will need to be addressed before the changes can be merged:
- The use of YAML has been changed to Yaml. This is fine if that is how it is referred to in Chinese. Otherwise the plugin will use "YAML".
- There are references to Latex, but the math block identifier is not a Latex parser, so I am not sure where Latex is coming from
- Tables are only recognized if they follow the guidelines set forth by github (Obsidian may allow other table types, but the Linter can only recognize github's tables)
- At least 1 blank line was added
- Why are the warning levels converted to all capitalized English values instead of using the Chinese words for those log levels?

pjkaufman (changes requested):
Thanks for putting in the work to help better the Chinese translation!

Most of the comments relate to the spelling of YAML so do not be worried. There are just about a dozen that are not.

I notice a couple of things that will need to be addressed before the changes can be merged:
- The use of YAML has been changed to Yaml. This is fine if that is how it is referred to in Chinese. Otherwise the plugin will use "YAML".
- There are references to Latex, but the math block identifier is not a Latex parser, so I am not sure where Latex is coming from
- Tables are only recognized if they follow the guidelines set forth by github (Obsidian may allow other table types, but the Linter can only recognize github's tables)
- At least 1 blank line was added
- Why are the warning levels converted to all capitalized English values instead of using the Chinese words for those log levels?

Ooopz:
Thanks for detailed review, some explain below:

1. I change YAML to Yaml just because it's a common usage of my company, both YAML or Yaml are fine. I can replace all Yaml to YAML insted.
2. Math block is highly related to Latxe, in Chinese Internet environment there is no different between math block and Latex block, this translate will be more clear to Chinese users.
3. Table is the same reason as Latex, github is redundant.
4. There are two blank lines, it should be my carelessness.
5. Warning levels do not need translate as I  do not translate `tags` and `aliases`, people use this features definitely konws the meaning. Another reason is we don't use Chinese tranlation to specialized term as debug level, user will feel weared when it be translated.

At final, I will replace all Yaml to YAML, and delete all blank lines. Latex and table condition will leave as it is.

Maybe a new pull request is easier to handle?

PR #803, opened 2023-07-14
chinese translate update

Ooopz:
Replace all Yaml to YAML, translate new added contents

### DIFF
--- a/src/lang/locale/zh-cn.ts
+++ b/src/lang/locale/zh-cn.ts
@@ -13,21 +13,21 @@ export default {
       'name': '格式化所有文件',
       'error-message': '格式化所有文件时出错',
       'success-message': '已格式化所有文件',
-      'errors-message-singular': '已格式化所有文件，但有 1 个错误',
-      'errors-message-plural': '已格式化所有文件，但有 {NUM} 个错误',
-      'start-message': '这将编辑所有文件，包括子文件夹中的文件，可能会引入错误',
+      'errors-message-singular': '已格式化所有文件，其中有 1 个错误',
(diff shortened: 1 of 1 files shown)


src/lang/locale/zh-cn.ts
```diff
@@ -740,44 +740,44 @@ export default {
...
-    'DEBUG': 'debug',
-    'INFO': '信息',
-    'WARN': '警告',
-    'SILENT': '静默',
+    'ERROR': 'ERROR',
+    'TRACE': 'TRACE',
+    'DEBUG': 'DEBUG',
+    'INFO': 'INFO',
+    'WARN': 'WARN',
+    'SILENT': 'SILENT',
```
pjkaufman, line 753: Why did these change to their English values instead of being their Chinese equivalents?

Ooopz: Just no need, programmers prefer to use the original English of programming terms, and I think debug level is only for programmers.

pjkaufman: These options are not just meant for developers as it is something that is supposed to be used when reporting issues as well.

pjkaufman: If that does not change anything, I am fine with merging this.

Ooopz: If there is any explanation of different debug level, I think translate these to Chinese will be meaningful, but we just get some words here, so keep it not translated is fine.

pjkaufman: Gocha. I will go ahead and merge this then.

PR #807, opened 2023-07-19
Fix Issue with Lint via File Menu Always Updating File Modified Date

### LINKED ISSUE #745: Bug: File Menu Lint Updates Date Modified in All Cases

pjkaufman:
Fixes #745 

Changes Made:
- Fix spacing around a comment
- Make sure to only call the method for modifying a file when the text has changed since for some reason that forces an update of the date modified when there is no change to the file contents itself

### DIFF
--- a/src/lang/locale/en.ts
+++ b/src/lang/locale/en.ts
@@ -169,7 +169,7 @@ export default {
       },
     },
     'debug': {
-    // debug-tab.ts
+      // debug-tab.ts
       'log-level': {
         'name': 'Log Level',
         'description': 'The types of logs that will be allowed to be logged by the service. The default is ERROR.',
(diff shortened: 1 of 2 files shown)


alitekdemir:
thanks

PR #838, opened 2023-08-02
Fix Linting Non-Markdown Files for `Lint on File Change`

### LINKED ISSUE #837: Bug: (serious) Linter lints pdf attachments

## Describe the Bug

With "Lint on File Change" on, after viewing a PDF in Obsidian 1.4.2, Linter lints the PDF and corrupts it.

To replicate. Choose a PDF that's an export of a Word document (True PDF, not a scanned PDF).

## Expected Behavior

I expect Linter to only lint .md files.

## Device

Windows 11 updated

## Additional Context

I don't know if it also happens when viewing an embeded PDF (likely not) but if you're fixing this please test that case too. Sorry I was too scared of what happened (This bug triggered others plugins to behave erratically, generating weird files, and crashed Obsidian).

I was really surprised that Linter works on PDF. Is there a way to limit Linter to work only on .md files?

pjkaufman:
Fixes #837 

Changes Made:
- Checked to make sure that the extension of the file is one of a markdown file

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -256,7 +256,7 @@ export default class LinterPlugin extends Plugin {
 
     const currentActiveFile = this.app.workspace.getActiveFile();
     const lastActiveFileExists = this.lastActiveFile == null ? false : await this.app.vault.adapter.exists(this.lastActiveFile.path);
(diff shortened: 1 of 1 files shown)


pjkaufman:
This still needs to be tested in Obsidian to verify it works.

pjkaufman:
This has been tested. All that is left is just to confirm that this works.

PR #854, opened 2023-08-19
Update Settings Components to Be More Uniform

### LINKED ISSUE #581: FR: Folder to ignore Examples and Functional Extensions Needed
An example of Folder to ignore is needed.
Also, it would be nice to set a folder that does not lint when

pjkaufman:
Fixes #581 

Made the logic for creating settings more uniform in the Linter. This should help out with making general settings that are not associated with a rule.

Changes Made:
- Converted log level from its int value to a string value to better work with other dropdowns
- Moved the CommonStyles type to where the settings data is since that is where it belongs
- Adding styling info for the copy option for the Linter's full text values on the debug tab to help people get a hold of their data.json and log contents
- Added a base setting for the general types of settings that are not associated with rules
- Added a generic dropdown option that handles enums and regular dropdown values
- Added a generic number input setting that should handle numeric value inputs
- Added logic to the full width text boxes for copying out their contents kind of like how github does
- Added a generic toggle setting that should handle boolean values for general settings
- Added 2 new svgs for the clipboard and success scenarios
- Replaced old settings with new implementations on general and debug tabs
- Added a way to add search info for the tabs using a generic setting to reduce the needed info to pass around
- Added a log levels enum and a way to take strings and set the values.
- Added a couple more helpers to nested key of for getting different types back
- Reworked how the custom row logic works and made it look more like the native Obsidian settings
- Revamped the folders to ignore logic to make it work just like custom commands for selecting folders


pjkaufman:
I am doing something a little different where I only update languages that I know, so the remainder will need to be updated/added by someone that already knows the language.

PR #891, opened 2023-09-14
Add british spelling for realise.

BradleyYeo:
How can I configure obsidian-linter to convert american spelling to british?

### DIFF
--- a/src/utils/auto-correct-misspellings.ts
+++ b/src/utils/auto-correct-misspellings.ts
@@ -1,4 +1,5 @@
 export const misspellingToCorrection = new Map<string, string>([
+  ['realize','realise'],
   ['1nd', '1st'],
   ['2rd', '2nd'],
   ['2st', '2nd'],


pjkaufman:
Hey @BradleyYeo , I am not sure the Linter is supposed to auto-correct for specofic dialects of English. You may just want to add that as a custom replacement value since `realize` is a valid word.

BradleyYeo:
Hi Peter,

Do you have suggestions for how I can do that?

Thank you,

Bradley

pjkaufman:
This is a little convoluted, but it is the only way the Linter currently allows for the functionality you are referring to:
You go to the custom tab in the Linter settings and have to add 2 rules per word you want to replace (one capitalized and one uncapitalized). So for what you are trying to do with realize you would add two rules to the custom replacement that would look something like the following:
image

pjkaufman:
There does seem to be a desire to extend the auto-correct list with custom values, but so far I am not sure as to the best way to go about it so I have left it alone. However I may go ahead and add something for it at some point. I just have a lot of things that have been requested and a lot of functionality to keep functional.

BradleyYeo:
It works thank you!

Thank you,

Bradley

pjkaufman:
I am going to go ahead and close this @BradleyYeo . Hopefully the example provided is enough to tackle your needs at this time.

PR #947, opened 2023-11-17
Fix: Linter break obsidian save file.

HananoshikaYomaru:
the previous save function is not called. Originally save function should trigger file to save immediately, but because of linter, file is not saved at all.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -228,15 +228,14 @@ export default class LinterPlugin extends Plugin {
       saveCommandDefinition.callback = () => {
         if (this.settings.lintOnSave && this.isEnabled) {
           const editor = this.getEditor();
-          if (!editor) {
-            return;
-          }
-
-          const file = this.app.workspace.getActiveFile();
-          if (!this.shouldIgnoreFile(file) && this.isMarkdownFile(file)) {
-            this.runLinterEditor(editor);
+          if (editor) {
+            const file = this.app.workspace.getActiveFile();
+            if (!this.shouldIgnoreFile(file) && this.isMarkdownFile(file)) {
+              this.runLinterEditor(editor);
+            }
           }
         }
+        save()
       };
     }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -55,6 +55,7 @@ export default class LinterPlugin extends Plugin {
   private lastActiveFile: TFile;
   private overridePaste: boolean = false;
   private customCommandsLock = new AsyncLock();
+  private originalSaveCallback?: () => void = null;
 
   async onload() {
     setLanguage(window.localStorage.getItem('language'));
@@ -85,6 +86,13 @@ export default class LinterPlugin extends Plugin {
 
     for (const eventRef of this.eventRefs) {
       this.app.workspace.offref(eventRef);
+    }
+
+    const saveCommandDefinition = this.app.commands?.commands?.[
+      'editor:save-file'
+    ];
+    if (saveCommandDefinition && saveCommandDefinition.callback && this.originalSaveCallback) {
+      saveCommandDefinition.callback = this.originalSaveCallback;
     }
   }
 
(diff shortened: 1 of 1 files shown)


HananoshikaYomaru:
@platers this is urgent so I guess you will want to check this out.

pjkaufman:
@HananoshikaYomaru , could you provide a reproducible scenario for this? I don't know why save would need to be called if you are not referring to an editor or a change was not made.

pjkaufman:
The save functionality is supposed to be hijacked when enabled however this should just be one of the many save callbacks from my understanding so I am not sure what the problem is so a reproducible example would help understanding what the problem is here.

HananoshikaYomaru:
I will make a video to describe when linter make vault on modify event delay

HananoshikaYomaru:
sorry I don't have time to make the video. 

but you can reproduce this using 

```ts
const saveCommandDefinition =
	this.app.commands?.commands?.["editor:save-file"];
const save = saveCommandDefinition?.callback;
if (typeof save === "function") {
	saveCommandDefinition.callback = () => {
		console.log(
			"user click save but since I hijacked it, save is not done"
		);
		save(); // you toggle comment of this and see the difference
	};
}

// defines the vim command for saving a file and lets the linter run on save for it
// accounts for https://github.com/platers/obsidian-linter/issues/19
const that = this;
window.CodeMirrorAdapter.commands.save = () => {
	that.app.commands.executeCommandById("editor:save-file");
};

this.registerEvent(
	this.app.vault.on("modify", (file) => {
		console.log("file is modified", file);
	})
);
```

HananoshikaYomaru:
this issue is usually not obvious but any plugin / workflow depends on modify event will have a wrong event firing time, other plugin would think user have saved the file but actually the files are not saved at all.

pjkaufman:
Given the information provided so far (no workflow that is broken or requires the file to save has been provided), I will need to look into this when I have some time, but this does not seem to be urgent. It will likely end up in the next release.

pjkaufman:
I will see about looking at what you showed. It looks like from discussing with other devs that this is an issue that I should be able to fix as you have suggested here. I am just going to verify that it works and once I do that (hopefully today), then this can be merged and go out in the next release. I am getting close to being ready for the next release, so it may be that this change would go out soon.

pjkaufman:
I am seeing an edge case that probably needs to be handled better:
- Each time the Linter is reloaded, the save command gets added again causing a lot of extra calls to save

To fix this, I am going to need to store a reference to the original save funciton on the plugin and set the save back to what it was prior to the Linter adding its values to prevent the duplication of the event over and over again.

pjkaufman:
Hey @HananoshikaYomaru , could you use the update I just made on the patch branch to see if that works for you? For some reason, I could not get the console log example to work for me in my test vault. I probably have configured something weird so I am not really able to reproduce the issue. If this change works for you, I am fine with going ahead and merging this.

HananoshikaYomaru:
https://github.com/platers/obsidian-linter/assets/43137033/760881df-4641-4130-ab06-6e54a83098da

@pjkaufman I made a video to show the PR is ready. Thanks man, appreciate your prompt reply.

pjkaufman:
No problem. Thanks for pointing out the problem!

HananoshikaYomaru:
https://www.youtube.com/watch?time_continue=422&v=u761OLGAI8M&embeds_referring_euri=https%3A%2F%2Fwww.redditmedia.com%2F&source_ve_path=Mjg2NjY&feature=emb_logo

Btw I create a custom save plugin which is a better way to handle this save function, I guess.

pjkaufman:
> https://www.youtube.com/watch?time_continue=422&v=u761OLGAI8M&embeds_referring_euri=https%3A%2F%2Fwww.redditmedia.com%2F&source_ve_path=Mjg2NjY&feature=emb_logo
> 
> Btw I create a custom save plugin which is a better way to handle this save function, I guess.

I am not sure that the logic needs to change in save logic. I do think it could be done better, but I only see 2 benefits that could be had potentially:
- Not needing to add back the save callback as it was prior
- Not using an undocumented callback

The errors thrown by the Linter should be caught by the Linter unless they happen outside of linting itself. So I am not 100% sure of the benefit of swapping how this works (especially since we still have to bind to the vim save logic anyways).

Please let me know if I am missing something. Thanks!

HananoshikaYomaru:
no worry, I am not asking for changing the code. I am just suggesting another way to deal with this issue.

PR #966, opened 2023-12-06
Custom Commands: Wait for Cache for Lint on File Change and Lint on Save

### LINKED ISSUE #827: FR: Beautified tables
Thanks for your great plugin.

I have some ugly tables, like this:

```md
| one | two  | three | four |
| --- | --- | --- | --- |
| 1 |✓| |
|2 | |✓| |
|3| |✓|✓|
| 4 | … | | someting else |
| Total | 100 | 20 | 300000000 |
```

Linter could change them to such a beautiful table:

```md
| one   | two | three | four          |
| ----- | --- | ----- | ------------- |
| 1     | ✓   |       |               |
| 2     |     | ✓     |               |
| 3     |     | ✓     | ✓             |
| 4     | …   |       | someting else |
| Total | 100 | 20    | 300000000     |
```

Ups - in this image you better see what I mean:

image

At them moment, I use these online pretifiers:

https://prettier.io/playground/
https://codebeautify.org/markdown-formatter

What do you think?

### LINKED ISSUE #930: FR: Run Commander hotkey on save / file change instead of linting

I'm using the [Advanced Tables](https://obsidian.md/plugins?id=table-editor-obsidian) plugin to format tables since Linter doesn't currently support this. To make the process smoother, I've used the [Commander](https://obsidian.md/plugins?id=cmdr) plugin to create a macro which triggers two hotkeys: The first formats all tables in the file (via Advanced Tables), and the second then lints 250 ms later (via Linter). This is great, but doesn't tie into on-save or on-file-change events that Linter tracks, requiring manual triggering of the Commander macro.

When Linter detects a save or file change, instead of triggering linting itself, it could allow a user to assign a hotkey to be triggered. This would let one trigger a hotkey created by Commander when a macro has been defined. Being able to trigger a hotkey (rather than tying into Commander specifically) would allow other plugins which may be for creating macros to also work.

If the feature was enabled, on file save or change, Linter would trigger the assigned hotkey and nothing more. The result in my case would be that my Markdown tables would get formatted, and then the file would get linted.

If Linter does anything time-specific, a big delay in a Commander macro could produce odd results. Linter should probably state that any unexpected behaviour as a result of triggering a hotkey on save/change is not Linter's problem but that of whichever plugin is responsible for the hotkey (e.g. Commander).

Manually trigger the hotkey when dealing with tables in Markdown.

Thanks.

pjkaufman:
Fixes #827 
Fixes #930 

There was an issue where some plugins were relying on the metadata cache to be up to date in order to work. This is normally true, but not right after the Linter has updated a file. Thus we needed to move the custom command running logic for single file lint into the metadata handler. It may be worthwhile to do the same for folder and vault lint to make sure that things work properly for them rather than relying on the sluggishness of those two kinds of linting mechanism to be enough for the cache to update like right now.

Changes Made::
- Updated custom commands to run in one of two ways after a file gets linted
  - The file contents change so we need to wait until the cache updates to run the custom commands
  - The file contents are the same so we should run the custom commands
- Added a couple of lists to keep track of files that recently were linted via editor save or via active file change
- Add a callback to the Linter for testing purposes that runs after the custom commands are called which should move to the next test that uses this setup automatically (has been tested with two test cases)


pjkaufman:
I need to double check why the Integration tests occasionally leave a second Linter tab present for the settings tabs.

pjkaufman:
> I need to double check why the Integration tests occasionally leave a second Linter tab present for the settings tabs.

I am not sure I will be able to fix this right now. I will likely need to put in an issue for this for the integration tests and tackle it down the road since this is really kicking my butt right now.

PR #970, opened 2023-12-09
Add helpUrl to manifest

TfTHacker:
I am adding helpUrl to the manifest to support the help system integration into Obsidian. Please consider supporting this initiative. Please note that this key value in manifest.json must be in the manifest downloaded as part of the plugin in the release.

https://forum.obsidian.md/t/links-to-help-and-manuals-for-plugins-and-themes/70733/8


pjkaufman:
This sounds like a good addition @TfTHacker . I just noticed the URL key in another manifest a couple of days ago.

I think adding it should be fine since the manifest should get auto added to the release when I create it.

TfTHacker:
Thank you and thank you for an amazing plugin.

PR #976, opened 2023-12-11
Potential Fix for Live Preview Properties Update

### LINKED ISSUE #973: Regression: Linting in Live Preview does not update metadata block display since 1.21.1
Unfortunately, it seems yet another regression made it into 1.21. This one also is not fixed by [the patch you posted here](https://github.com/platers/obsidian-linter/issues/969#issuecomment-1848847291).

---

## Describe the Bug / How to Reproduce

Hard to explain, probably better showcased with a screen recording:

https://github.com/platers/obsidian-linter/assets/73286100/acd0e508-525b-4670-ab51-3e39bb0588e8

- In my config, I have the linting rule enabled that adds the modification date as `mdate`.
- When in source view, it updates correctly.
- However, when in Live Preview or Reading Mode, linting does not update the *display* of the metadata block. As can be seen via the notification and by switching to source mode, the `mdate` is in fact inserted correctly. It is only the display of the metadata block that is not updated correctly.
- This behavior did not occur before v. 1.21.0.

pjkaufman:
Fixes #973 

There seems to be an issue caused by calling the original save callback right after linting a file. This causes Live Preview and Reading View's properties view to not update immediately. I am not sure what causes this, but likely the save event causes a refreshing of the cache or something that prevents the immediate UI update. So it seems best to do the original save callback and then do the Linter logic. This could cause 2 save events, but the debounce should prevent that from my understanding.

Changes Made:
- Made the original save callback run before linting the file

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -246,6 +246,8 @@ export default class LinterPlugin extends Plugin {
 
     if (typeof this.originalSaveCallback === 'function') {
       saveCommandDefinition.callback = () => {
+        this.originalSaveCallback();
+
         if (this.settings.lintOnSave && this.isEnabled) {
           const editor = this.getEditor();
           if (editor) {
@@ -255,8 +257,6 @@ export default class LinterPlugin extends Plugin {
             }
           }
         }
-
-        this.originalSaveCallback();
       };
     }
 


pjkaufman:
Based on the information I have at this time, I believe this is a good fix for the problem. We can come back and change things up if it causes some kind of issue.

PR #979, opened 2023-12-11
Rework Footnote Logic

### LINKED ISSUE #902: Bug: Re-Index footnotes also moves footnotes to the bottom

## Describe the Bug

*Re-Index Footnotes* moves footnotes to the bottom, even when *Move Footnotes to the bottom* is unticked.

## How to Reproduce

1. In Linter settings, make sure *Re-index Footnotes* is turned on and *Move Footnotes to the bottom* is turned off.
2. Make a new document with the following text (without blockquotes):

```markdown
foo [^1]

[^1]: bar

foobar
```

3. Lint manually or via auto-lint on save.

## Expected Behavior

Expected behavior is that nothing should happen.

## Device

- [x] Desktop
- [x] Mobile (iOS)

pjkaufman:
Fixes #902 

There was an issue where re-index footnotes was moving them to the end of the document. This was caused by trying to sort the references by their location in the document which is handled by the rule for moving footnotes to the end of the document. Thus that logic was removed. Re-index footnotes was then updated to make sure it properly handled removing duplicate footnote definitions.

Changes Made:
- Moved the test for sorting index values to move footnotes to end of file where it belongs since re-index should not be moving footnotes to the end of the file and move footnotes already does the sorting
- Added test for making sure that footnote definitions are left where they are in the file
- Duplicate footnote definitions are removed where they are at
  - For inline definitions that means they are removed in place
  - Definitions on their own line are removed with their line
-  Added some expect error info for some typescript warnings around mdast and fixed a function call that was not properly setup
- Cleaned up and condensed the footnote related logic to make it simpler and work more like what is expected

### DIFF
--- a/__tests__/move-footnotes-to-the-bottom.test.ts
+++ b/__tests__/move-footnotes-to-the-bottom.test.ts
@@ -233,5 +233,46 @@ ruleTest({
         [^5]: f-footnote
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/902
+      testName: 'Moving footnotes to the end of the file should make sure that there are no empty lines between footnotes',
+      before: dedent`
+        AAA[^1]
+        ${''}
+        BBB[^2]
+        ${''}
+        Something[^foo-bar]
+        ${''}
+        CCC[^hello-world]
+        ${''}
+        DDD[^3]
+        ${''}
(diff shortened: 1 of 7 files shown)


pjkaufman:
I think this looks good especially with the feedback I got. I am going to merge this and we can iron out any other issues with it as they arise.

Issue #954, opened 2023-11-29
FR: Format File Names


Ken9969:
Although I have encountered some situations, I believe what I would like to see added is mostly self-explanatory.

1) I would like Linter to analyze the tags portion of the YAML front matter and delete any duplicate tags. This situation occurs when using Tag Wrangler to delete tags. As Tag Wrangler does not and will not in the future delete tags, the workaround is to rename unwanted tags to something that exists and is wanted. Whenever I pick an existing tag, if that tag happens to already exist in that particular document, it becomes duplicated. In my Obsidian, because of the size, that happens a lot. Because there are many ways to invoke tags and because some folk may not want to invoke such a feature, I would suggest it as an option.

2) Linter does a fine job of formatting headings to the title case. I would like to see this extended to encompass the file title. Again, it could be made optional.

Regards,

Ken

pjkaufman:
Hey @Ken9969 . I believe duplicate tag removal has already been requested as #857 (though this is more specific than that).

As for the file title, could you explain some more? The file title is not Markdown, so it is a bit weird to add into the list of Linter rules. However it may be feasible to add depending upon what the logic in question is.

Ken9969:
Thank you, Peter. I didn't realize this had been requested earlier. I did read your reference #857. That post refers to aliases, not tags. There is an earlier request referring to tags, #508. Still, I don't think I should delete it because it offers a bit of explanation.

Concerning your second paragraph, I must report that I do not understand. My tags suggestion does refer to Markdown files that I have in Obsidian. Could you possibly give me a better explanation or maybe an example of the change you suggest?

pjkaufman:
> Thank you, Peter. I didn't realize this had been requested earlier. I did read your reference #857. That post refers to aliases, not tags. There is an earlier request referring to tags, #508. Still, I don't think I should delete it because it offers a bit of explanation.

I have no qualms with keeping this open at the same time. It is slightly different, but I believe the title refers to adding a rule to dedupe values in all frontmatter arrays. So I thought it might overlap with the first request. So leaving this open is fine. Just wanted to confirm there was overlap so if the other gets implemented this could be closed as well (assuming the other part gets satisfied or rejected).

> Concerning your second paragraph, I must report that I do not understand. My tags suggestion does refer to Markdown files that I have in Obsidian. Could you possibly give me a better explanation or maybe an example of the change you suggest?

Sorry, I did not explain that well. My point about the file title is in reference to
[...]

Issue #508, opened 2022-11-20
FR: Sort and De-Duplicate Frontmatter Tags


grantx2016:
## Is Your Feature Request Related to a Problem? Please Describe.

After migrating a gazillion notes into Obsidian from another app, and in the course of ongoing tag manipulation using vs code and regex, I am finding an overwhelming number of tag duplicates appearing in the frontmatter.

## Describe the Solution You'd Like

Sort the multiline frontmatter tag list alphabetically and then de-duplicate it to leave a clean and more readable list of tags.

**Please include an example where applicable:**

``` markdown

---
tags:
  - 25-ict/org/test2
  - 17-marketing/test1
  - 25-ict/org/test1
  - 25-ict/org/test2
  - general
  - 24-human-resources/people
  - 17-marketing/test1
  - 25-ict/org/test3
  - 24-human-resources/people
---
```

``` markdown
---
tags:
  - 17-marketing/test1
  - 24-human-resources/people
  - 25-ict/org/test1
  - 25-ict/org/test2
  - 25-ict/org/test3
  - general
---
```

## Describe Alternatives You've Considered

I have toyed with writing a separate regex script using awk to do this but, apart from my limited skills in this area, considered it to be a good fit for this linter plugin, and if implemented, would be a more robust solution that would benefit others.

## Additional Context

This feature would also ensure that the tag count is correct as it would no longer be counting duplicates.

P.S. Keep up the great work on this plugin! One of my favorites. 👍

pjkaufman:
A change has been merged that adds the logic to remove duplicate YAML array values. It should go out in the next release. Please let us know if that is not the case.

pjkaufman:
I just went ahead and wrote something for sorting the YAML array values. It does not give you the granularity to decide different sort orders for different array types (tags, aliases, normal array). But if we want to, that can be added later.

pjkaufman:
It has been merged into master and should go out with the next release. Please let us know if this is not the case.

PR #985, opened 2023-12-19
Dedupe YAML Arrays

### LINKED ISSUE #857: FR: Delete Duplicate Array Values from YAML
Further to our discussion at #855, please add a toggle or perhaps preferably a rule (without a setting) that removes any duplicate keys, e.g. from alias property (actually any number over one instance should be removed).  

Of course, diactric characters like `ä`, etc. are not to be taken as equals of `a`, etc. And "Lion" does not equal "lion".  But these are of course self-explanatory.

Thank you,

Best

pjkaufman:
Fixes #857 
Relates to #954 
Relates to #508 

Adds the ability to remove duplicate values from a YAML array.

Changes Made:
- Add a new rule for removing duplicate YAML array values
- Tries to abide by the current YAML format as much as possible
- Adds options to enable duplicate removal for tags, aliases, normal arrays, and an option to list out keys of YAML arrays to ignore


grantx2016:
I just tested this with the latest release of Linter and it's not working for me. Similarly the sorting not working either. I can confirm that I have enabled these options in the plugin settings. By comparison, if I manually select the relevant lines in the YAML block and run the command 'sort u' then it dedupes and sorts correctly.

pjkaufman:
I just pushed a patch. Please check version 1.23.1.

Drakemoor:
It works for me now. Thank you.

grantx2016:
All working A-OK. Thanks for the quick update for this feature. It's a ripper. 👍

Issue #927, opened 2023-10-27
Bug: I noticed that Obsidian was freezing on the trait text.


wangyu92:
- [ X ] I have verified that I am on the latest version of the Linter

## Describe the Bug

I was using Obsidian and noticed that my app kept freezing. I started by disabling all plugins and found that the freezing was caused by the Linter plugin. I also found that this happened all the time under certain circumstances.

## How to Reproduce

<img width="522" alt="image" src="https://github.com/platers/obsidian-linter/assets/20921042/ed7dd3a0-3d3b-4d12-a8eb-b63d7a88bbe5">

Example to reproduce issue with

``` markdown
---
tags: []
title: Bug Reproduce
date created: 2023-10-28T02:29:30
date modified: 2023-10-28T02:30:36
---
- Hi. This is a note to reproduce the bug.
- There must be at least one bullet point.

- Add an empty line and write one or more bullet points again like this.
- Then run the linter plugin and the Obsidian app should freeze.
```

## Expected Behavior

In short, your Obsidian app will freeze.

## Device

- [x] Desktop: Silicon mac
- [ ] Mobile (I haven't tried it on mobile.)

pjkaufman:
Could you prived your `data.json`, @wangyu92 ? I know there are some situations that the Linter does not handle the best.

pjkaufman:
Hey @wangyu92 , could you provide your `data.json`? Without it, I may not be able to address this potential issue.

pjkaufman:
I am guessing this is caused by the Linter not using workers to Lint files. However without your `data.json` @wangyu92 , I will not be able to look into exactly what is causing the problem since I cannot reproduce locally with just the info provided.

wangyu92:
I really sorry for too late reply. I upload `data.json` file.

[data.json](https://github.com/platers/obsidian-linter/files/13331794/data.json)

pjkaufman:
Sorry about the delay in responding. Could you try out the following `main.zip` by unzipping its contents and copying the `main.js` file to `<vault_base>/.obsidian/plugins/obsidian-linter`? This could have been related to the Live Preview issues since I see that you are using live preview.

Here is the `main.zip`: 
[main.zip](https://github.com/platers/obsidian-linter/files/13648953/main.zip)

If you are not able to test it manually, the change for Live Preview should go out with the next release.

pjkaufman:
[...]

Issue #872, opened 2023-09-01
Bug: Linter freezes Obsidian V1.4.5 on Fedora Silverblue 38 Flatpak


dougkeiller:
- [ ] I have verified that I am on the latest version of the Linter

## Describe the Bug

When Linter "lints" a file, Obsidian freezes. Only way to unfreeze is to restart.

Disabling Linter prevents this from happening, so definitely narrowed to Linter as cause of freeze.

## Rules

If you are not sure which rule(s) may cause the issue, please attach the `data.json` file located at `<vault>/.obsidian/plugins/obsidian-linter/` or copy the config from the debug tab.

{
  "ruleConfigs": {
    "escape-yaml-special-characters": {
      "enabled": false,
      "try-to-escape-single-line-arrays": false
    },
    "force-yaml-escape": {
      "enabled": false,
      "force-yaml-escape-keys": ""
    },
    "format-tags-in-yaml": {
      "enabled": false
    },
    "format-yaml-array": {
      "enabled": false,
      "alias-key": true,
      "tag-key": true,
      "default-array-style": "single-line",
      "default-array-keys": true,
      "force-single-line-array-style": "",
      "force-multi-line-array-style": ""
    },
    "insert-yaml-attributes": {
      "enabled": false,
      "text-to-insert": "aliases: \ntags: "
    },
    "move-tags-to-yaml": {
      "enabled": false,
      "how-to-handle-existing-tags": "Nothing",
      "tags-to-ignore": ""
    },
    "remove-yaml-keys": {
      "enabled": false,
      "yaml-keys-to-remove": ""
    },
    "yaml-key-sort": {
      "enabled": false,
      "yaml-key-priority-sort-order": "",
      "priority-keys-at-start-of-yaml": true,
      "yaml-sort-order-for-other-keys": "None"
    },
    "yaml-timestamp": {
      "enabled": false,
      "date-created": true,
      "date-created-key": "date created",
      "force-retention-of-create-value": false,
      "date-modified": true,
      "date-modified-key": "date modified",
      "format": "dddd, MMMM Do YYYY, h:mm:ss a"
    },
    "yaml-title": {
      "enabled": false,
      "title-key": "title",
      "mode": "first-h1-or-filename-if-h1-missing"
    },
    "yaml-title-alias": {
      "enabled": false,
      "preserve-existing-alias-section-style": true,
      "keep-alias-that-matches-the-filename": false,
[...]

PR #1047, opened 2024-03-16
Draft: Move Most Linting Logic to Web Workers

### LINKED ISSUE #337: FR: Use Web Workers to Improve Lint All and Lint Folder Experience
Currently when lint all is run, it runs asyncronously, which is fine so long as there are not a lot of files to parse. I have somewhere around 2k and it tends to lock up Obsidian when I run lint all to test out new rules. It would also be nice to have a persisted notice of the progress like what dataview does on load.

So the two requests are to offload linting to Web Workers where more than 1 is being linted at a time. The other is to allow the user to be aware of the amount of files found that will be linted versus the amount completed so far.

pjkaufman:
Fixes #337 
Relates to #927 
Relates to #872 

To make the linting experience better for users, it is better if we can run the linting logic on a separate thread from the main thread. This will allow the user to be able to work on things if linting a file is taking a while. Most rules can run in the web workers. But the ones that deal with moment or obsidian specific logic have to remain on the main thread since they cannot have their data passed to and from the web worker (this is a limitation due to the complexity of the moment object) and because the source code is not available at compile time which makes it impossible to inline the logic from obsidian. So we have to compromise a little here. But this should still be a net improvement for users with larger files or who use mass linting options on their vaults. This will need to be in BETA for some time if possible. So I will need to setup BRAT for this change.

TODO:
- [ ] Swap out uses of the `RulesRunner` class for references to the lint file manager
- [ ] Make sure linting on active file change works
- [ ] Make sure linting on active file change works when custom commands are used and no change is made
- [ ] Make sure linting on active file change works when custom commands are used and a change is made
- [ ] Make sure regular linting works
- [ ] Make sure regular linting works when custom commands are used and no change is made
- [ ] Make sure regular linting works when custom commands are used and a change is made
- [ ] Make sure linting a folder works
- [ ] Make sure linting the vault works
- [ ] Make sure linting on paste works
- [ ] Make sure that linting a file with YAML frontmatter that says that all rules are disabled does not run custom commands
- [ ] Make sure that regular rule disabling via the YAML frontmatter is respected
- [ ] Remove old copy folder and files
- [ ] Move linting functions to the `utils` folder as a `lint` or `lint-runner` file since it is just a set of related functions that are useful when running linting operations
  - These should mostly be used by the file lint manager, but we shall see
- [ ] See about setting a timeout on how long the worker can run before it will throw an error to prevent an infinite loop from happening and not allowing more linting to happen
  - I am thinking that 5 or 10 minutes should be a good value to use as a measurement
  - This will help out with making sure that we prevent long running lints from happening unexpectedly
  - We may want a user override, but that can probably wait until users request it

If we want to have more than 1 worker running at a time:
- [ ] See about adding file name to logs in console log for lints since there may be 2+ workers now which will make things pretty hectic when it comes to making sure that logs are displayed back to back in the console


pjkaufman:
I am going to beef up the integration test suite for this to make sure I have a higher degree of confidence that this is working as expected. Right now I am not entirely sure that I am at a good point to move forward with this PR since I am feeling a little overwhelmed thinking about all of the manual testing that will be needed. If I add enough ITs, it should give me a better feeling of certainty as to the reliability of these changes.

pjkaufman:
I am going to start on this again and just try to get it ready for beta testing. I plan to focus on this unless really easy or really pressing issues pop up in the meantime.

pjkaufman:
I believe linting a file and folder are working. The log situation has yet to be ironed out. But it looks like changes are being made. I am still only testing using 1 web worker. When this goes live, I would like to have it be around 3.

PR #1070, opened 2024-04-06
Basic implementation of YAML timestamp UTC option

### LINKED ISSUE #1069: FR: Add option to save YAML timestamps in UTC
I'm currently using my Obsidian vault in conjunction with external tooling (TiddlyWiki) that expects the created & modified YAML frontmatter datetimes to always be in UTC. Adding an option to save YAML timestamps in UTC allows Linter's timestamps to be interoperable with TiddlyWiki and other external tools that might only be able to work with UTC.

I'd like the option to choose whether or not local times are converted to UTC times before being formatted & saved in created / modified frontmatter. For my specific use case, a boolean option is sufficient.

For example, if a file was created at 9 PM Eastern Standard Time on Jan 1st 2020 and convertToUTC was true, the file's _date created_ value would be 2 AM Jan 2nd 2020 in UTC.

Local time:
``` markdown
---
date created: 2020-01-01T21:00:00-05:00
---
```
Converted to UTC:
``` markdown
---
date created: 2020-01-02T02:00:00+00:00
---
```

Re: possible alternative solutions - I've attempted to modify TiddlyWiki to work with non-UTC time stamps and it's not feasible for me at this time. This also seems outside the scope of a change that would make sense for Obsidian's core. Adding to Linter makes the most sense to me for wider compatibility.

lucible:
Fixes #1069.

This is my rough attempt at implementing a option to convert YAML timestamps from local time to UTC time. I've added a few tests, but I'm unsure of their coverage / if any additional ones are needed. Happy to take any feedback and rework my implementation if desired.

When creation date retention is enabled and the format string does not include time zone info, this implementation assumes the existing creation date is already in UTC (no change). If the format string does include time zone info, the existing creation date is in a local time, and the UTC option is subsequently enabled, then the existing date created will be converted to UTC on next lint.


pjkaufman:
Just as a heads up, making changes to this PR will not allow you to create a BRAT release. In order to make a BRAT release you would need to be able to create a tag or release on the main repo. If you want this locally until it is merged into master, you will need to build it locally and install it manually.

pjkaufman:
I hopefully will be able to look into this PR later this week.

lucible:
Apologies, I didn't realize those changes for building locally would get pulled in on this PR! I've reverted those so this branch only includes the relevant code changes to this PR.

I also noticed in testing that the yaml-keysort rules use currentTime, so added a commit that updates rules-runner to check for the UTC option.

pjkaufman:
I will fix the conflict I created and see about testing the logic here come Wednesday. Sorry about the delay on looking at this.

pjkaufman:
As an update, I was not able to get to this Wednesday, so I will have to take a look at this tomorrow.

pjkaufman (approved):
LGTM

pjkaufman:
Thanks for working on this @lucible! It looks good to me. I am going to go ahead and merge this.

Issue #1112, opened 2024-06-29
Bug: footnote-after-punctuation fails if reference name contains hyphen


domodwyer:
## Describe the Bug

With `footnote-after-punctuation` enabled:

* `Some text[^test].` correctly becomes `Some text.[^test]`.

but

* `Some text[^test-me].` is incorrectly left with the footnote before the punctuation.

## How to Reproduce

Create a file with:

```text
Some text[^test-me].
Some text[^test].
```

And then lint this file with `footnote-after-punctuation` enabled.

<details><summary>Debug Log</summary>
<p>

```text
Running linter
rules before regular rules: 0.10000014305114746 ms
Running Add Blank Line After YAML
add-blank-line-after-yaml: 0.2999999523162842 ms

Some text[^test-me].
Some text[^test].
Running Dedupe YAML Array Values
dedupe-yaml-array-values: 0.40000009536743164 ms

Some text[^test-me].
Some text[^test].
Running Move Tags to YAML
move-tags-to-yaml: 1.2000000476837158 ms

Some text[^test-me].
Some text[^test].
Running Header Increment
header-increment: 0.10000014305114746 ms

Some text[^test-me].
Some text[^test].
Running Headings Start Line
headings-start-line: 0.10000014305114746 ms

Some text[^test-me].
Some text[^test].
Running Remove Trailing Punctuation in Heading
remove-trailing-punctuation-in-heading: 0.09999990463256836 ms

Some text[^test-me].
Some text[^test].
Running Footnote after Punctuation
footnote-after-punctuation: 0 ms

Some text[^test-me].
Some text.[^test]
Running Convert Bullet List Markers
convert-bullet-list-markers: 0.40000009536743164 ms

Some text[^test-me].
Some text.[^test]
Running Emphasis Style
emphasis-style: 0.20000004768371582 ms

Some text[^test-me].
Some text.[^test]
Running No Bare URLs
no-bare-urls: 0.09999990463256836 ms

Some text[^test-me].
Some text.[^test]
Running Remove Empty List Markers
remove-empty-list-markers: 0.3000001907348633 ms

Some text[^test-me].
Some text.[^test]
Running Remove Hyphenated Line Breaks
remove-hyphenated-line-breaks: 0.09999990463256836 ms

Some text[^test-me].
Some text.[^test]
Running Remove Multiple Spaces
remove-multiple-spaces: 0.2999999523162842 ms

Some text[^test-me].
Some text.[^test]
Running Strong Style
[...]

PR #1113, opened 2024-06-29
fix: footnote-after-punctuation containing hyphen

domodwyer:
This allows symbols to appear in footnote references without breaking the `footnote-after-punctuation` lint.

The old regex used `\w` so it only matched references that consisted of words, digits and whitespaces - I swapped it to accept anything except newlines and the reference terminating `]` character, so now it accepts all symbols too.

I'm not sure how to include tests for this but I'm happy to if you can point me in the right direction :+1:

Fixes https://github.com/platers/obsidian-linter/issues/1112.

---

* fix: footnote-after-punctuation containing hyphen (10b7dac)
      
      Fixes the footnote-after-punctuation rewriting for footnote references
      that contain hyphens (or other symbols):
      
        * Bananas[^test].
        * Bananas[^test-me].
        * Bananas[^test.2].
      
      Is now rewrote to:
      
        * Bananas.[^test]
        * Bananas.[^test-me]   <-- fixed
        * Bananas.[^test.2]    <-- fixed
      
      Fixes https://github.com/platers/obsidian-linter/issues/1112.

### DIFF
--- a/__tests__/footnote-after-punctuation.test.ts
+++ b/__tests__/footnote-after-punctuation.test.ts
@@ -1,11 +1,11 @@
 import dedent from 'ts-dedent';
 import FootnoteAfterPunctuation from '../src/rules/footnote-after-punctuation';
-import { ruleTest } from './common';
+import {ruleTest} from './common';
 
 ruleTest({
   RuleBuilderClass: FootnoteAfterPunctuation,
   testCases: [
-        { // fixes https://github.com/platers/obsidian-linter/issues/1112
+    { // fixes https://github.com/platers/obsidian-linter/issues/1112
       testName: 'Simple case',
       before: dedent`
         Some text[^test-me].
@@ -18,5 +18,5 @@ ruleTest({
         Some text.[^test]
       `,
     },
-],
+  ],
 });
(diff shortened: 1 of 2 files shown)


pjkaufman:
Hey @domodwyer , this looks pretty good as is. But I think we should at least add a test for this scenario to the code. It should be feasible to add a test for the scenario you are trying to catch by adding a file like the one located [here](https://github.com/platers/obsidian-linter/blob/master/__tests__/move-footnotes-to-the-bottom.test.ts).

So you would create a file similar to it called `footnote-after-punctuation.test.ts`. Then you would add something like the following to it:
``` TypeScript
import FootnoteAfterPunctuation from '../src/rules/footnote-after-punctuation';
import dedent from 'ts-dedent';
import {ruleTest} from './common';

ruleTest({
  RuleBuilderClass: FootnoteAfterPunctuation,
  testCases: [
        { // fixes https://github.com/platers/obsidian-linter/issues/1112
      testName: 'Simple case',
      before: dedent`
        Some text[^test-me].
        Some text[^test].
      `,
      after: dedent`
        Some text.[^test-me]
        Some text.[^test]
      `,
    },
],
});
```

domodwyer:
Ha, I don't know how I missed the folder named `__tests__`! Thanks for the helpful pointers @pjkaufman.

I've pushed 5d86f4d that adds the test file - I haven't run it locally (typescript isn't my usual sandbox!) so I'll keep an eye on CI and fix up any issues :+1:

Thanks for the help!

domodwyer:
[Success!](https://github.com/platers/obsidian-linter/actions/runs/9757399897/job/26929708126?pr=1113#step:6:83)

This should be good to go now 🙏

pjkaufman:
Thanks for putting in the work to add that functionality @domodwyer . It should go out in the next release.

PR #1122, opened 2024-07-11
Support Obsidian-style comment for inline linter toggle

### LINKED ISSUE #1121: FR: Support Obsidian-style comment for inline linter toggle

Writing HTML comments (`<!-- -->`), aside from being tedious normally, is especially annoying when using software or plugins that attempt to auto-correct successive hyphens and other characters into distinct typographical marks. For example, [Smart Typography](https://github.com/mgmeyers/obsidian-smart-typography) ([view in Obsidian](https://r.0xyz.com/?to=obsidian%3A%2F%2Fshow-plugin%3Fid%3Dobsidian-smart-typography)) makes typing an HTML comment look something like this:
`<!----` *`<BS>`* `linter-disable----` *`<BS>`* `>` *`<BS>`*

Is that not a problem of this plugin? Absolutely. But only looking at HTML comments ignores Obsidian's commenting syntax which could be used as an alternative, for more reasons than just this one.

Allow for `%%linter-(en|dis)able%%` as an alternative.

Use templates instead of writing out the comment.

## Additional Context

#1122

sevmonster:
Fixes #1121

### DIFF
--- a/__tests__/get-all-custom-ignore-sections-in-text.test.ts
+++ b/__tests__/get-all-custom-ignore-sections-in-text.test.ts
@@ -29,6 +29,16 @@ const getCustomIgnoreSectionsInTextTestCases: customIgnoresInTextTestCase[] = [
     expectedPositions: [],
   },
   {
+    name: 'when no custom ignore start indicator is present, no positions are returned even if custom ignore end indicator is present when Obsidian comment format is used',
+    text: dedent`
+      Here is some text
+      %% linter-enable %%
+      Here is some more text
+    `,
+    expectedCustomIgnoresInText: 0,
+    expectedPositions: [],
+  },
+  {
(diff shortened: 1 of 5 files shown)


pjkaufman:
I think the only concerns I have right now are that no UTs were added and that the docs would need updating to include the syntax reference to this new syntax.

The three tests shown by [this search](https://github.com/search?q=repo%3Aplaters%2Fobsidian-linter+%3C%21--+language%3ATypeScript&type=code&l=TypeScript) should have a duplicated test for the new syntax. I can update the docs or if you want to the source for the docs in question is located [here](https://github.com/platers/obsidian-linter/blob/2e5aeee102321dcb3c311a63a0648f658c44e5aa/docs/docs/usage/disabling-rules.md#L52-L73).

pjkaufman:
I went ahead and made the two requested additions. Assuming UTs pass, I will go ahead and merge this.

PR #1171, opened 2024-09-22
horizontal rule spacing

### LINKED ISSUE #1170: FR: Empty line around horizontal rules

I would like the same functionality that adds empty lines to headings, blockquotes, etc but for horizontal rules. 

I use horizontal rules to separate thoughts in the same way as headings but without providing a title. Linter can add empty lines around headings and I would like the same empty lines around horizontal rules.

I only want it to add spaces around actual horizontal rules not triple dashes that are used for frontmatter.

**Example:**

``` markdown
---
prop: value
---

# h1
asdf
---
qwer
```

``` markdown
---
prop: value
---

# h1

asdf

---

qwer
```

I tried to come with regex to do it but after a few failed attempts by chatgpt o1 I gave up.

thomas-forbes:
Fixes #1170 

I can add more tests when I get the green light about adding this general functionality.

### DIFF
--- a/__tests__/empty-line-around-horizontal-rules.test.ts
+++ b/__tests__/empty-line-around-horizontal-rules.test.ts
@@ -7,6 +7,129 @@ ruleTest({
   testCases: [
     {
       testName:
+      'Horizontal rules that start a document do get an empty line before them.',
+      before: dedent`
+        ***
+        ${''}
+        ${''}
+        Content
+      `,
+      after: dedent`
+        ***
+        ${''}
+        Content
+      `,
+    },
+    {
+      testName: 'Horizontal rules that end a document do not get an empty line after them.',
+      before: dedent`
+        ***
+        Content
+        ***
+      `,
+      after: dedent`
+        ***
+        ${''}
+        Content
+        ${''}
+        ***
+      `,
+    },
+    {
+      testName: 'YAML frontmatter is not affected by this rule',
+      before: dedent`
+        ---
+        prop: value
+        ---
+        ${''}
+        Content
+      `,
+      after: dedent`
+        ---
+        prop: value
+        ---
+        ${''}
+        Content
+      `,
+    },
+    {
+      testName: 'All types of horizontal rules are affected by this rule',
+      before: dedent`
+        - Content 1
+        ***
+        - Content 2
+        ---
+        - Content 3
+        ___
+        - Content 4
+      `,
+      after: dedent`
+        - Content 1
+        ${''}
+        ***
+        ${''}
+        - Content 2
+        ${''}
+        ---
+        ${''}
+        - Content 3
+        ${''}
+        ___
+        ${''}
+        - Content 4
+      `,
+    },
+    {
+      testName: 'Paragraphs above `---` are treated as a heading and not spaced apart',
+      before: dedent`
+        Content
+        ---
+      `,
+      after: dedent`
+        Content
+        ---
+      `,
+    },
+    {
+      testName: 'Horizontal rules between blockquotes',
+      before: dedent`
+        > Blockquote
+        > > Nested Blockquote
+        ---
+        > Blockquote
+        > > Nested Blockquote
+      `,
+      after: dedent`
+        > Blockquote
+        > > Nested Blockquote
+        ${''}
+        ---
+        ${''}
+        > Blockquote
+        > > Nested Blockquote
+      `,
+    },
+    {
+      testName: 'Horizontal rules inside blockquotes',
+      before: dedent`
+        > Blockquote
+        ---
+        > > Nested Blockquote
+        > Blockquote
+        > > Nested Blockquote
+      `,
+      after: dedent`
+        > Blockquote
+        ${''}
+        ---
+        ${''}
+        > > Nested Blockquote
+        > Blockquote
+        > > Nested Blockquote
+      `,
+    },
+    {
+      testName:
         'Make sure that basic dash horizontal rules have lines added around them',
       before: dedent`
         - asdf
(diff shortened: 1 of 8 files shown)


pjkaufman:
Do you mind clarifying the need for the addition of frontmatter parsing? I assume it is because `---` can be a horizontal rule, but I would like to double check.

thomas-forbes:
Yes, without the frontmatter package the AST recognizes the `---` in the frontmatter definition as thematic break but with the package it recognizes the `---` as a part of the frontmatter. eg

```markdown
---
prop: value
---

asdf
--- 
qwer
```

pjkaufman:
By chance does that happen when the YAML is ignored for that rule? I forget what exactly is used as the ignore value, but I think it is `---\n---`.

pjkaufman:
> Sorry I don't understand the question. Are you asking whether we would get the same behavior if we ignored yaml for this rule? Is there a way to do that without the frontmatter package? I didn't know about that but that should work.

There is for the Linter. [It uses a field to determine what types are to be ignored for the rule](https://github.com/platers/obsidian-linter/blob/cc23589d778fb56b95fe53b499e9f35683a2b129/src/rules/convert-spaces-to-tabs.ts#L17). The ignore type for YAML is defined [here](https://github.com/platers/obsidian-linter/blob/cc23589d778fb56b95fe53b499e9f35683a2b129/src/utils/ignore-types.ts#L24). It would allow for changing the YAML to `---\n---`. So the above input would need to start with that instead. So you would add that to the ignore list if it does indeed keep from getting recognized as a horizontal rule.

pjkaufman:
I just did a quick check and it looks like it is still necessary to have the frontmatter parser present otherwise it will incorrectly parse things. I will have to take a closer look at this PR later this week and give more feedback then.

src/rules/empty-line-around-horizontal-rules.ts
```diff
@@ -0,0 +1,51 @@
...
+          ***
+          ${''}
+          asdf
+          ${''}
+          ___
+          ${''}
+          qwer
+        `,
+      }),
+    ];
```
pjkaufman, line 46: We will want to have at least two more cases present here:
- One where the hr is at the end of the document
- One where the hr is in the middle of the document
- Preferably one that shows that YAML is not affected as well

src/utils/mdast.ts
```diff
@@ -68,10 +70,11 @@ function parseTextToAST(text: string): Root {
 
   // @ts-expect-error for some reason an overload is missing
   const ast = fromMarkdown(text, {
-    extensions: [combineExtensions([gfmFootnote(), gfmTaskListItem()]), math()],
+    extensions: [combineExtensions([gfmFootnote(), gfmTaskListItem(), frontmatter(['yaml', 'toml'])]), math()],
```
pjkaufman, line 73: Why pass in `toml` here? I would prefer we only use `yaml`.

src/utils/mdast.ts
```diff
@@ -68,10 +70,11 @@ function parseTextToAST(text: string): Root {
...
     mdastExtensions: [[
       gfmFootnoteFromMarkdown(),
       gfmTaskListItemFromMarkdown,
+      frontmatterFromMarkdown(['yaml', 'toml']),
```
pjkaufman, line 77: Why pass in `toml` here? I would prefer we only use `yaml`.

__tests__/empty-line-around-horizontal-rules.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        prop: value
+        ---
+        ${''}
+        asdf
+        ${''}
+        ___
+        ${''}
+        qwer
+      `,
+    },
```
pjkaufman, line 76: We will want to add some corresponding tests that test how the hrs work in blockquotes and nested blockquotes to make sure that the blank line generation works as intended for those scenarios as well.

pjkaufman (changes requested):
So overall this looks pretty good. I want to thank you for taking the time to add this.

A couple of things to note about the PR:
- I mentioned a couple of UTs that I would expect to see on the PR
- I also added some comments around removing the use of `toml` in the parser as I don't believe it is indeed needed
- I also wanted to say that the wording for the new rule looks good. But all of the other changes to the localization file will need to be reverted as that does not seem to comport with the style that the Linter has been using up until now (I am not sure if something needs to be added to the Linter configs to help whatever editor you are using respect the Linter's eslint and other rules)

__tests__/empty-line-around-horizontal-rules.test.ts
thomas-forbes: Is what I added what you're looking for?

thomas-forbes:
Ok I have made the changes you requested. Let me know if there is anything else you think needs to be changed

__tests__/empty-line-around-horizontal-rules.test.ts
pjkaufman: This is definitely close. With the blockquote scenario, I was more wondering about the following scenarios:
Assuming horizontal rules can be in a blockquote:
``` markdown
> Blockquote
>
> ---
> > Nested Blockquote
> Blockquote
> > Nested Blockquote
```
And the following edge cases (how does it handle blockquotes of the same level with an empty blockquote line between them):
``` markdown
> > Blockquote
> > 
> > ---
>
> > Nested Blockquote
> Blockquote
> > Nested Blockquote

Callout scenario:
> > Blockquote
> > 
> > ---
>
> > [!Info] Nested Blockquote
> > Info here
> Blockquote
> > Nested Blockquote

```

pjkaufman (commented):
Other than the two edge cases/scenarios I mentioned. This looks good.

pjkaufman:
I think I am going to merge this even with the two edge cases not accounted for as they are pretty big edge cases that have come up in the past, but typically do not come up very frequently. So I think we are fine without them for now.

PR #1175, opened 2024-09-27
Feat: `Auto-Correct Common Mispellings` Allow Users to Ignore Words with Multiple Capitals

### LINKED ISSUE #1075: FR: Don't auto-complete acronyms

For example, ['aadd', 'add'], converts aadd to add regardless of capitals. If AaDD is an acronym, it gets converted to Add.

Add an option to not auto-complete if characters after the first character is capital.

I would still like aadd and Aadd to be converted to add and Add.
Just not things like AADD, aADD, aaDD, aadD, AADd, AAdD, etc.

Cant think of any alternatives

## Additional Context

n/a

### LINKED ISSUE #1166: FR: Auto-correct should ignore words in all caps (aka acronyms)

I was writing up some notes about insurance benefits and wrote `HSA` which is a perfectly valid acronym but the rule changed this to `HAS`. This is not a good correction. I had to add `hsa` as a word to ignore but in normal writing I _would_ want that to get replaced with `has`.

I think that words that are only uppercase should be ignored. Or at least there should be an option to toggle this. I think larger scale the autocorrect _should_ be case sensitive in general but that seems less likely to be added.

I've used the [Auto Correct](https://marketplace.visualstudio.com/items?itemName=sygene.auto-correct) VSCode extension for a long time and have always enjoyed how it handled things and wish this rule worked like that does.

pjkaufman:
Fixes #1075 
Fixes #1166 

There were two requests to allow for auto-correct to ignore capital letters. This is mostly to be used to help make sure acronyms are ignored while still allowing non-acronyms for the same words (i.e. hsa to has vs. HSA) to be properly handled.

Changes Made:
- Added option to turn off auto-correct for words with multiple capital letters in them
- Added UTs to check that the rule in question does indeed work

### DIFF
--- a/src/lang/locale/en.ts
+++ b/src/lang/locale/en.ts
@@ -265,6 +265,10 @@ export default {
         'name': 'Extra Auto-Correct Source Files',
         'description': 'These are files that have a markdown table in them that have the initial word and the word to correct it to (these are case insensitive corrections). **Note: the tables used should have the starting and ending `|` indicators present for each line.**',
       },
+      'skip-words-with-multiple-capitals': {
+        'name': 'Skip Words with Multiple Capitals',
+        'description': 'Will skip any files that have a capital letter in them other than as the first letter of the word, so acronyms and some other words can benefit from this. It may cause issues with proper nouns being properly fixed.',
+      },
     },
     // add-blank-line-after-yaml.ts
     'add-blank-line-after-yaml': {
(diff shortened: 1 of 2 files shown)


Kainkun:
Thanks!

pjkaufman:
No problem. Hopefully this helps out with a shortcoming of the rule which was that it was only ever designed to be used with lowercase words/generate lowercase words minus the first latter.

PR #1178, opened 2024-09-28
Feat: Update Date Modified on Editor Content Change & Miscellaneous Style and YAML Timestamp Changes

### LINKED ISSUE #183: FR: Auto run the linter after X minutes since last edit
Add a option to run this plugin automatically after no more edit were made to a file after X minutes.

This will enable the updated / created timestamps to be always up to date, it's easy to forget to manually save a file, and thus, not having the keys updates

### LINKED ISSUE #890: FR: Get modified date property behave like created date property
The created date property has the option "Force Date Created Key Value Retention" which is useful to retain created date when moving the vault around. Unfortunately the "Date modified" property does not have that option and thus the only practical way to keep it from updating when moving the vault around is to use another plugin "Update frontmatter modified date" by Alan Grainger. 

Hope the problem is clear, [here](https://forum.obsidian.md/t/how-to-add-an-automatically-updated-last-edit-time-in-document-properties/67224/4?u=ironhak) is a slightly better explaination if needed.

### LINKED ISSUE #995: FR: Set created property to current date

I think the `created` property is being populated with the file creation time (`crtime`) from the filesystem metadata (file attrs). Some file systems, like btrfs, don't use `crtime` and linter defaults the `created` property to _1970-01-01_.

When the `created` property is initially populated, if it is _1970-01-01_, set it to the current system time instead.

Current behavior on save of new note:

``` markdown
---
created: 1970-01-01
modified: 2024-01-16
---
```

Proposed behavior on save of new note:

``` markdown
---
created: 2024-01-16
modified: 2024-01-16
---
```

Manually updating the field ... but that would defeat the purpose of having a tool do that automatically.

## Additional Context

Nil.

### LINKED ISSUE #1161: FR: Lint on file change isn't checking whether the file was actually changed

I use Linter to set the `date_modified` field in my YAML so that I can keep track of what files I've been working on each day with a dataview query. However the "Lint on File Change" option is a bit misleading, since it doesn't detect whether I've actually changed a file, it just lints when I close the file.

Is it possible to have some kind of check to see whether I've actually modified the file and only then run the linter?

_N/A_

I've tried using the inbuilt metadata (`file.mday`) to make the dataview queries but this fails when I use Sync across devices since it'll return different things depending on which computer I'm using. The Linter "Lint on File Change" seems to be the best solution if it can actually detect the file change.

## Additional Context

_N/A_

### LINKED ISSUE #1162: Bug: YAML modified timestamp / order of operations
- [x] (1.26.0.rc3)

## Describe the Bug

The order of operations is wrong, leading to unnecessary / excessive "modified" YAML property updates.

## How to Reproduce

1. Create a note with a YAML "modified" timestamp property set to be auto updated by Linter.

2. Fill the note with typical markdown (text, lists, etc).

3. Make change(s) `A`, that conflict with *any* other configured Linter rule. Examples: double blank line, a new bullet point with no text, a double space between words.

On save, Linter will make the correction(s). Yet despite the note content being precisely the same as before change `A`, the YAML timestamp of the designated "modified" property will still be updated.

## Expected Behavior

Linter should be executing all the linting operations *first*, then doing a diff of the file to determine whether the content has *actually* changed. Then finally updating the YAML "modified" property if necessary.

### Reasoning

Human error is inevitable. The entire design and purpose of linters/formatters, is acknowledging this fact to preempt it.

But in this specific case, Linters solution is *ignoring human error* and plastering over the modified property to "make it pretty".

I realize this proposal will lead to inconsistency with meta-data tracked by the file system. But aside from it maybe triggering my OCD a little, so what?

It's not like when the commit date / merge date in git doesn't line up precisely with the modified file attribute(s), devs are like: `"Welp it must be invalid, i gotta mod the commit / revert..."`

Why should this be any different?

I'd much rather have the modified timestamp be *accurate and meaningful* in relation to the content itself, over "the formality" of displaying filesystem / OS related data.

And in the scenario users want "the formality", it's not as though there aren't other options in Obsidian:

- https://github.com/chetachiezikeuzor/Get-Info-Plugin
- https://github.com/Yustynn/obsidian-last-modified-timestamp-in-status-bar
- https://github.com/paultreanor/notes-dater

### Side benefit

**Disclaimer**: I haven't looked deeply into Linters source, so from here it might just be me waffling.

The simplest way to diff the file contents is to store an additional copy of it in memory ie. File as first opened in an editor vs file after lint ops are complete.

There is also another aspect of the codebase that uses this functionality. Debugging and Testing.

pjkaufman:
Fixes #183 
Fixes #890 
Fixes #1161 
Fixes #1162 
May fix #995

There has been some level of desire for the ability to make the date modified be set whenever the user makes a change to the editor/file via Obsidian. In order to do this, there is a need to know whenever the editor is updated. When it is, it will set off a debounced action to update the date modified so long as the user does not remove the changes by the time that the debounced function runs.

There were also some style changes made to cut down some on the whitespace that was present, but not necessary.

There was also a wording update to try to make things clearer that the lint on file change only happens when focused file changes.

Changes Made:
- Updated `Lint on File Change` to `Lint on Focused File Change`
- Adds ability to run the YAML timestamp rule x seconds after the last editor change (it tries to keep from updating the date modified if no net changes were made in the editor between when the first change was made and the debounce ran)
- Tries to make sure that the other linting logic at the end updates the original file text for any debounce functions that exist to try to keep from unnecessarily adding a another timestamp update if it is not necessary
- Adds a dropdown for determining the source of truth for date modified and date created

TODO:
- [x] Swap boolean setting for trusting created date in the YAML as the source of truth to a dropdown
- [x] Add a similar option for date modified where only changes in Obsidian are recorded
- [x] Try to account for no valid created date provided (0 as the ctime should instead be now)

### DIFF
--- a/__integration__/obsidian-mode.test.ts
+++ b/__integration__/obsidian-mode.test.ts
@@ -37,7 +37,7 @@ function edgeCaseSetup(plugin: TestLinterPlugin, _: Editor): Promise<void> {
     'enabled': true,
     'date-created': true,
     'date-created-key': 'created',
-    'force-retention-of-create-value': true,
+    'date-created-source-of-truth': 'file system',
     'date-modified': true,
     'date-modified-key': 'last_modified',
     'format': 'YYYY-MM-DD',
--- a/__tests__/yaml-timestamp.test.ts
+++ b/__tests__/yaml-timestamp.test.ts
@@ -63,12 +63,12 @@ ruleTest({
         format: 'dddd, MMMM Do YYYY, h:mm a ',
         currentTime: moment('Thursday, January 2nd 2020, 12:01 am', 'dddd, MMMM Do YYYY, h:mm a'),
         alreadyModified: false,
-        forceRetentionOfCreatedValue: false,
+        dateCreatedSourceOfTruth: 'frontmatter',
         fileModifiedTime: '2020-01-02T00:00:00-00',
       },
     },
     {
-      testName: 'When the date format changes and `forceRetentionOfCreatedValue = true`, date created value is based on the one in the YAML frontmatter.',
+      testName: 'When the date format changes and `dateCreatedSourceOfTruth = frontmatter`, date created value is based on the one in the YAML frontmatter.',
       before: dedent`
         ---
         created: Wednesday, January 1st 2020, 12:00:00 am
@@ -82,7 +82,7 @@ ruleTest({
       options: {
         dateModified: false,
         dateCreatedKey: 'created',
-        forceRetentionOfCreatedValue: true,
+        dateCreatedSourceOfTruth: 'frontmatter',
         format: 'YYYY, h:mm:ss a',
         locale: 'en',
       },
@@ -359,7 +359,7 @@ ruleTest({
       },
     },
     { // accounts for https://github.com/platers/obsidian-linter/issues/745
-      testName: 'When no changes are made, and force retention of creation date is active, do not update date modified when no change in modification time has been made',
+      testName: 'When no changes are made, and frontmatter is the source of truth, do not update date modified when no change in modification time has been made',
       before: dedent`
         ---
         tag: tag1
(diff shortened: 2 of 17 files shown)


baodrate:
`lintOnFileContentChangeDelay` doesn't appear to have a corresponding selector in the settings UI. is this intentional?

PR #1185, opened 2024-10-07
Move Default Auto-Correct Misspellings to Separate File

pjkaufman:
There has been a push to make plugins more performant for the Obsidian app. The `main.js` for the Linter is about 1.4MB in size. Most of the bundled size (.9MB) comes from the default misspellings. However this is really only needed if auto-correct common misspellings is enabled. I suspect that not all users are using this rule, but it is causing the bundle to be so large that it is slowing down the load time of the plugin which thus impacts the load time of the user's vault. To help speed this up, it is being moved to its own Markdown file. This Markdown file should be installed once when the rule is enabled or if it is missing. This also allows for the auto-correct file to be loaded when the layout is ready. This also greatly helps improve the loading time of the plugin.

The zip file generated as part of release should now include the default Markdown file to help manual installations work well.


pjkaufman:
I need to update where it is pointing for the default misspellings list on the settings page, but after that this should be clear to be merged.

Issue #1217, opened 2024-11-11
Bug: The `Dedupe YAML Array Values` rule fails to detect differently escaped variations of the same value as duplicates


rxlecky:
- [x] I have verified that I am on the latest version of the Linter

## Describe the Bug

The `Dedupe YAML Array Values` rule fails to detect differently escaped variations of the same value as duplicates.

## How to Reproduce

Here's an example markdown file with a duplicate value in its array property, only differing in how the value variations are escaped:

``` markdown
---
array: [a, "a", 'a']
---

# Untitled

... note text here ...
```

## Expected Behavior
When running linter with `dedupe-yaml-array-values` rule turned on, I would expect the three instances of the value `a` to be collapsed into one. Instead, the property is left unchanged.

## Config
```
{
  "ruleConfigs": {
    "add-blank-line-after-yaml": {
      "enabled": false
    },
    "dedupe-yaml-array-values": {
      "enabled": true,
      "dedupe-alias-key": true,
      "dedupe-tag-key": true,
      "dedupe-array-keys": true,
      "ignore-keys": ""
    },
    "escape-yaml-special-characters": {
      "enabled": false,
      "try-to-escape-single-line-arrays": true
    },
    "force-yaml-escape": {
      "enabled": false,
      "force-yaml-escape-keys": ""
    },
    "format-tags-in-yaml": {
      "enabled": false
    },
    "format-yaml-array": {
      "enabled": false,
      "alias-key": true,
      "tag-key": true,
      "default-array-style": "single-line",
      "default-array-keys": true,
      "force-single-line-array-style": "",
      "force-multi-line-array-style": ""
    },
    "insert-yaml-attributes": {
      "enabled": false,
      "text-to-insert": "aliases: \ntags: "
    },
    "move-tags-to-yaml": {
      "enabled": false,
      "how-to-handle-existing-tags": "Nothing",
      "tags-to-ignore": ""
    },
    "remove-yaml-keys": {
      "enabled": false,
      "yaml-keys-to-remove": ""
    },
    "sort-yaml-array-values": {
      "enabled": false,
      "sort-alias-key": true,
      "sort-tag-key": true,
      "sort-array-keys": true,
      "ignore-keys": "",
      "sort-order": "Ascending Alphabetical"
    },
    "yaml-key-sort": {
      "enabled": false,
      "yaml-key-priority-sort-order": "",
      "priority-keys-at-start-of-yaml": true,
[...]

Issue #992, opened 2024-01-07
Bug: YAML Key Sort Does Not Support Line Break Syntax


bryanwhiting:
- [x] I have verified that I am on the latest version of the Linter

## Describe the Bug

Linting with line breaks doesn't work properly with the sorting feature. You'll notice the description falls down into the "draft" property. 

Before lint:

```
title: 'hello'
description: |
  "Something nut"
date: 2024-01-07T00:00:00
updated: 2024-01-07T12:18:20
categories: favorite
draft: false
hi: |
  "Hello world"
bye: |
  There bob
``` 

After lint:

```
---
title: 'hello'
description: |
date: 2024-01-07T00:00:00
updated: 2024-01-07T12:21:18
categories: favorite
draft: false
  "Something nut"
hi: |
  "Hello world"
bye: |
  There bob
---
```

## How to Reproduce

Sorting yaml rules in options 

``` 
title:
description:
date:
updated:
categories:
draft:
``` 

## Expected Behavior

It should not shift the stuff after "|" down. 

## Screenshots

See above 

## Device

- [ ] Desktop
- [x] Mobile

pjkaufman:
Hey @bryanwhiting . It looks like you ran into a
some YAML syntax that the Linter does not know how to handle. I am not sure when or if this will get addressed since there currently is not a good way to handle this scenario without reformatting the key value pairs at this time.

If someone would like to add a fix for this or handle this syntax before I or someone else does, feel free to make a PR.

bryanwhiting:
Thanks for the quick reply. I'll just turn off sorting for now then, as I'm probably useless in trying to resolve this. 

Turning off sorting keys prevents the problem.

pjkaufman:
I have merged a change that should fix this. It may cause other issues, so I plan to put this out as beta release that will hopefully help us determine any issues that exist prior to having it roll out to all users.

Issue #912, opened 2023-10-10
Bug: YAML Key Sort doesn't support some YAML syntax


redactedscribe:
- [x] I have verified that I am on the latest version of the Linter

Apologies if this issue has already been created. Not quite sure what to search for.

## Describe the Bug

According to https://www.yamllint.com/, this is valid:

```yaml
---
AAA:
  - BBB: x
    CCC: x
  - DDD: x
    EEE: x
    FFF:
      - GGG: x
      - HHH
    III: x
FFF:
  -
```

`FFF` is present twice, but at different levels, and therefore should be unique, but obsidian-linter still complains with (even if line with `GGG` has `: x` removed):

```
plugin:obsidian-linter:43 [Obsidian Linter] Lint File Error in File 'Test.md':
Error "YAML Key Sort" encountered an error in the YAML:  bad indentation of a sequence entry (35:8)

 32 |     FFF:
 33 |   - GGG: x
 34 |   - HHH
 35 |     III: x
-------------^
 36 | FFF:
 37 |   - 
YAMLException: bad indentation of a sequence entry (35:8)

 32 |     FFF:
 33 |   - GGG: x
 34 |   - HHH
 35 |     III: x
-------------^
 36 | FFF:
 37 |   - 
    at Cc (plugin:obsidian-linter:18:944)
    at K (plugin:obsidian-linter:18:978)
    at dc (plugin:obsidian-linter:25:696)
    at Wt (plugin:obsidian-linter:25:4861)
    at Og (plugin:obsidian-linter:25:2150)
    at Wt (plugin:obsidian-linter:25:4870)
    at Ig (plugin:obsidian-linter:25:7247)
    at Mc (plugin:obsidian-linter:26:258)
    at Fg (plugin:obsidian-linter:26:469)
    at Mi (plugin:obsidian-linter:104:223)
Uc @ plugin:obsidian-linter:43
```

## How to Reproduce

1. Insert the front matter above into a file.
2. Lint with YAML Key Sort enabled (screenshot below for settings).

## Expected Behavior

Support this YAML syntax.

## Screenshots

![Obsidian_2023-10-11_00-21-34](https://github.com/platers/obsidian-linter/assets/25013746/5fd549b7-c613-425a-aba0-1c09de1d163a)

## Device

- [x] Desktop
- [ ] Mobile

Thanks.

redactedscribe:
Other observations:

[...]

Issue #715, opened 2023-05-10
Bug: Removes front matter list of arrays when there is a comment.


android272:
## Describe the Bug

List of arrays is removed from Yaml frontmatter when there is a comment. I thought the liter was removing the list of arrays at first but then for some reason I manually removed the comment and the list of arrays did not get removed. I don't see `comments` anywhere in the settings so maybe the ability to keep/remove comments in frontmatter shold be added.

## How to Reproduce

This section of front matter is added to my recipe notes

``` markdown
---
Ingredients: # [`<free text for ingredient>`,`<amount>`, `<unit>`, `<form>`, `<note>`]
  - [Watermelon, 80 g, cubed, note]
  - [sugar, 20 g]
---
```

No logs with `error` set.

```
{
  "ruleConfigs": {
    "escape-yaml-special-characters": {
      "enabled": true,
      "try-to-escape-single-line-arrays": true
    },
    "force-yaml-escape": {
      "enabled": false,
      "force-yaml-escape-keys": ""
    },
    "format-tags-in-yaml": {
      "enabled": true
    },
    "format-yaml-array": {
      "enabled": true,
      "alias-key": true,
      "tag-key": true,
      "default-array-style": "single-line",
      "default-array-keys": true,
      "force-single-line-array-style": "",
      "force-multi-line-array-style": ""
    },
    "insert-yaml-attributes": {
      "enabled": true,
      "text-to-insert": "Aliases: []"
    },
    "move-tags-to-yaml": {
      "enabled": false,
      "how-to-handle-existing-tags": "Nothing",
      "tags-to-ignore": "Prayer\nGratitude/"
    },
    "remove-yaml-keys": {
      "enabled": false,
      "yaml-keys-to-remove": "Date\nTime"
    },
    "yaml-key-sort": {
      "enabled": true,
      "yaml-key-priority-sort-order": "",
      "priority-keys-at-start-of-yaml": true,
      "yaml-sort-order-for-other-keys": "Ascending Alphabetical"
    },
    "yaml-timestamp": {
      "enabled": true,
      "date-created": true,
      "date-created-key": "date created",
      "force-retention-of-create-value": false,
      "date-modified": true,
      "date-modified-key": "date modified",
      "format": "YYYY-MM-DD, h:mm:ss a"
    },
    "yaml-title": {
      "enabled": false,
      "title-key": "title"
    },
    "yaml-title-alias": {
      "enabled": false,
      "preserve-existing-alias-section-style": true,
[...]

PR #1227, opened 2024-11-22
Swap to from `js-yaml` to `yaml` for Parsing YAML

### LINKED ISSUE #1082: Bug: YAML Key Sort option removes multi-line content after literal operator `|`

## Describe the Bug

When the YAML Key Sort option `Sorts the YAML keys based on the order and priority specified.` and a `YAML Sort Order for Other Keys` order  is enabled, nested content after a literal operator `|` is removed.

This is a problem for me because I'm using [Obsidian Custom Sort](https://github.com/SebastianMC/obsidian-custom-sort), which recommends syntax with a `sorting-spec` key followed by a literal operator.

## How to Reproduce

1. Enable `Settings > Linter > Sorts the YAML keys based on the order and priority specified.`
2. Set `YAML Sort Order for Other Keys` to `Ascending Alphabetical` or `Descending Alphabetical`.
3. Add front matter with a literal operator and nested content.
4. Run the linter.
5. The nested lines after the `|` will be removed.

### Example

```yaml
---
sorting-spec: |
  order-desc: a-z
---
```

`order-desc: a-z` will be removed.

## Expected Behavior

Nested content after a literal operator should not be removed.

## Device

- [X] Desktop
- [ ] Mobile

pjkaufman:
Fixes #1082 #912 #992 #660 #632 #649 #1258

The following are a list of tickets that are potentially fixable with the merger of this change, but logic will need to change in order to properly handle these scenarios:

Relates to #715 #1217 #617

The main reason for making this change is getting closer to having a round trip parser for the Linter to use on the YAML. This change helps with supporting more YAML syntax as well as having a less buggy experience for end users. It should result in fewer instances where things get broken.

It should do its best to keep things as they were, but there is no guarantee that that will be the case for all formats. However it will do its best given that it is using the CST instead of the AST for serializing to YAML.


pjkaufman:
If I am understanding correctly, the regex still works when 
``` yaml
FFF:
  - 
```
becomes
```yaml
FFF:
  
  - 
```
So I think it should be fine to proceed with this change. I may want to add some more changes first to see how things go.

pjkaufman:
I checked again to make sure and it seems it converts the value from 
``` yaml
FFF:
  - 
```
to
```yaml
FFF:
  
  - 
```
to
``` yaml
FFF:

  
  - 
```

which the regex does not support. So it is not actually feasible to swap over until all of the regex calls are fixed. However this also brings in some doubt that the underlying parser also knows that the ending version is the original version of the content.

pjkaufman:
Looks like that was a bug that was present prior. It has been fixed after I reported the bug to the library.

pjkaufman:
I would like to swap over to using the CST for more granularity. One place it is currently used is https://github.com/apollographql/argocd-config-updater/blob/0217630ad16edf17d8ff98239cca3c4b29426ec9/src/yaml.ts

pjkaufman:
I believe the CST is now being properly used to help retain the YAML as closely to what it was originally. This does not currently do anything for the regex used to grab values and move them anywhere except in the yaml-key-sort logic right now.

PR #1232, opened 2024-11-26
add new lint rule: align table columns

Ooopz:
new lint rule, test locally several times, and I have no idea why unit test failed, maybe you can fix it.


pjkaufman:
Hey @Ooopz . Thanks for taking the time to put in a PR. If I read correctly what the UT is talking about, there is a missing newline in the expected string.

I am a little hesitant to merge this on the basis that I am not sure if this works on CJK, Monospace, and non monospace fonts. I am not sure what the expectations are here, but often times the alignment of table columns fails one or more of these scenarios.

Ooopz:
In the align logic , the lengths of Wide and Fullwidth characters (including CJK characters) are set to 2, while the lengths of other characters are set to 1. So in some monospaced fonts, which widths ratio of English characters to CJK characters is 1:2,  the alignment results are good. In non monospaced fonts and some monospaced fonts widths ratio not be 1:2, the final rendered results are not perfect but acceptable. However, the final display effect depends on the fonts used, and I think in terms of code we can't do any further.

Ooopz:
Perhaps a reminder can be added in the settings that the alignment may not perform perfectly under non-monospaced fonts？

PR #1279, opened 2025-02-23
chore: refactor runLinterFile to use vault.process for more safety

kdnk:
Use `this.app.vault.process` for more safe file modifications.
ref. https://youtu.be/RvtV9DKr3ag?si=ch9cn9T0HMlNtoGg&t=52
ref. https://docs.obsidian.md/Plugins/Vault#Modify+files

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1021,7 +1021,7 @@ export default class LinterPlugin extends Plugin {
     await this.customCommandsLock.acquire('command', async () => {
       this.currentlyOpeningSidebar = true;
 
-      await sidebarTab.openFile(file);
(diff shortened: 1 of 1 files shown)


pjkaufman:
Hey @kdnk , sorry about the delay in responding to this PR. I do like the idea of doing this. And it is likely something we should do. I am just not 100% on board with doing this right now since I am off and on working on getting the logic to run in an async way. This would mean that we would have to get rid of the `vault.process` logic. This is not to say I will not merge this. I am just not decided yet on whether or not to merge the PR.

Hopefully that makes sense while I think this over.

PR #1286, opened 2025-03-07
Minor change: Only show notice with actual changes

Moyf:
To avoid the "0 character added, 0 character removed" notice

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -818,7 +818,11 @@ export default class LinterPlugin extends Plugin {
         ${charsAdded} ${getTextInLanguage('notice-text.characters-added')}
         ${charsRemoved} ${getTextInLanguage('notice-text.characters-removed')}
       `;
-      new Notice(message);
+
+      // Only show the notification if there are actual changes
+      if (charsRemoved + charsAdded > 0) {
+        new Notice(message);
+      }
     }
   }
 


pjkaufman:
Hey @Moyf , thanks for creating the PR. I am a little torn on whether to merge it or not. I may have the wrong use case here for why I would keep the notice since I use it to let me know if things were linted and nothing happened. But maybe that is just something I use it for that most users do not want.

I am thinking about this and hope to come to a decision on it.

Moyf:
> Hey @Moyf , thanks for creating the PR. I am a little torn on whether to merge it or not. I may have the wrong use case here for why I would keep the notice since I use it to let me know if things were linted and nothing happened. But maybe that is just something I use it for that most users do not want.
> 
> I am thinking about this and hope to come to a decision on it.

Oh hey! I think this PR is actually related to how I personally use Linter—I bind "Linter formatting" and "saving" together (and assign it to `Ctrl+S`), so my trigger frequency might be significantly higher than that of a typical user (who manually performs Lint operations).  
For cases where Lint operations are performed manually, yes, even "0 changes" is necessary feedback to let them know that their command has indeed been executed.  

Perhaps (if you don't mind adding further settings) "**Do not show Notice when modifications are empty**" could be made an optional setting (defaulting to False)?  

> Also, strangely, I seem to recall that there was a setting toggle for "whether to show Notice" itself, but I just tried to find it and couldn't... Did I misremember? 🤔

pjkaufman:
@Moyf , that setting addition would be fine with me. Just note that the setting would go under the general tab and would need to only show if the Linter has enabled the notice for the changes made. There are some similar examples in that file here:  https://github.com/platers/obsidian-linter/blob/162f0ac7753890661d95cad9b50c1b0cad22b897/src/ui/linter-components/tab-components/general-tab.ts#L39-L55

pjkaufman:
I believe there are two notices. There is one for lint on file change and there is one for lint on save. But then again I could be wrong and need to double check that logic.

Moyf:
> I believe there are two notices. There is one for lint on file change and there is one for lint on save. But then again I could be wrong and need to double check that logic.

Yes, you're right, there are two options for notice.
image

What confused me before is that even when I turn off the `Lint on save` option (I use the **Commander** plugin to manually bind `Save`, `Lint` and some of my scripts together), the option still works.

In other words, that option is not *only* applied to the `Lint on save` option; even when I manually use `Lint the current file` command, it still follows this setting to show notice.
So I think it shouldn't only display when the `Lint on save` option is enabled. I wonder if my understanding is correct? 🤔

pjkaufman:
Yes, the setting should actually not be hidden by Lint on Save. I incorrectly added the hiding logic.

As for how I think that the new setting would work, I think it would show if either display message on lint or display lint on file change message is enabled. That way it can be a global setting on the notices without needing multiple versions of the settings.

Moyf:
> Yes, the setting should actually not be hidden by Lint on Save. I incorrectly added the hiding logic.
> 
> As for how I think that the new setting would work, I think it would show if either display message on lint or display lint on file change message is enabled. That way it can be a global setting on the notices without needing multiple versions of the settings.

Yes, I agree with you 🫡
When I got time, I'll add the new changes to this PR   : )

Moyf:
> Yes, the setting should actually not be hidden by Lint on Save. I incorrectly added the hiding logic.
> 
> As for how I think that the new setting would work, I think it would show if either display message on lint or display lint on file change message is enabled. That way it can be a global setting on the notices without needing multiple versions of the settings.

@pjkaufman Hi, long time no see! I've finally completed the version of this PR with settings, opened at #1378.

PR #1306, opened 2025-05-03
fix: correct regex pattern for title matching in yaml-title rule

nshcr:
The yaml-title rule adds a title key to the frontmatter or replaces it if it already exists.
I noticed that if other keys start with the title key (e.g., titleTemplate, one of the VitePress's frontmatter config), they might be incorrectly replaced — especially if they appear before title.
So I updated the regex to match the exact title key and avoid this issue.
Thanks for reviewing.

### DIFF
--- a/__tests__/yaml-title.test.ts
+++ b/__tests__/yaml-title.test.ts
@@ -357,5 +357,22 @@ ruleTest({
         # [test \#4](github.com/platers/obsidian-linter)
       `,
     },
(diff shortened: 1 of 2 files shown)


pjkaufman:
@nshcr , this looks good to me. Thanks for taking the time to fix this. I will go ahead and merge this since it looks to be passing UTs and does not look like it would cause any issues.

PR #1307, opened 2025-05-03
fix typo: `twp-space-line-break` to `two-space-line-break`

nshcr:


pjkaufman:
Hey @nshcr , this seems fine to me. The only real issue I see is that the user's data.json values are not getting mapped accordingly, so this could cause users to lose their setting data. Here is one such example of moving the settings data in the current logic: https://github.com/platers/obsidian-linter/blob/b4dd3d9f05a393d4765e8823c4d8f8b380dcaae4/src/main.ts#L636-L642

Would you want to add that logic to this PR?

pjkaufman:
This looks good @nshcr . I will go ahead and merge this. Thanks for making the change.

PR #1326, opened 2025-05-27
Auto-release 1.29.2

pjkaufman:


redactedscribe:
@pjkaufman This .2 release doesn't seem to be available through Obsidian as I still haven't been offered it. I'm on 1.29.1. Do I need to manually install it, and if so, will it interfere with receiving future updates?

pjkaufman:
@redactedscribe , are you on at least version 1.9.0? If not, you will need to revert down to version 1.29.0. There was a bad version added to the versions file at one point and BRAT seems to have been accidentally not validating the minimum version of Obsidian. Those two things likely caused the ability to install an incorrect version of the Linter if that is what is going on for you.

redactedscribe:
@pjkaufman Sorry, 1.9.0? My Obsidian is 1.8.10, but that's not what you mean, is it? I'll see if I can revert back to 1.29.0.

pjkaufman:
@redactedscribe , to revert the version it will depend on how you install the Linter. If you use BRAT, the Linter will need to be updated to have a static version for the time being and specify the version to be `1.29.0`. Otherwise, I believe you have to manually install the plugin version from the release: https://github.com/platers/obsidian-linter/releases/tag/1.29.0

redactedscribe:
@pjkaufman I backed up my data.json before doing anything just in case. I was under the wrong assumption that by downgrading Linter Obsidian would then offer me version 1.29.2, but as far as I can see, this version is simply not available via Obsidian. It took me a while to understand this, but 1.29.0 is working fine. I don't think I've ever used BRAT; I'll upgrade when a new version is available via Obsidian.

Why hasn't the 1.29.2 "Lint on Save" bugfix version made it to Obsidian yet? There are probably quite a few people on 1.29.1 experiencing this bug.

Also, another thing I've noticed in case it's something which has been overlooked: The 1.29.0 ZIP download via GitHub contains "default-misspellings.md" whereas 1.29.0 installed via Obsidian does not.

pjkaufman:
@redactedscribe , the  obsidian version 1.9.0 introduce a change in how they setup the save callback. This broke the logic for the insider build. So version 1.29.1 of the Linter was released to address that. However it had a bug which was then fixed in version 1.29.2.

If you are not yet on 1.9.0, you have no need of the `Lint on Save` fix from either version 1.29.1 or 1.29.2. It will actually break `Lint on Save` if you try using either version prior to Obsidian 1.9.0

redactedscribe:
@pjkaufman I'm not on 1.9.0 because Obsidian 1.9.0 doesn't exist? Latest from obsidian.md is 1.8.10: 

https://obsidian.md/download: https://github.com/obsidianmd/obsidian-releases/releases/download/v1.8.10/Obsidian-1.8.10.exe

pjkaufman:
@redactedscribe , Obsidian 1.9.0 is an insider build at this time.

redactedscribe:
@pjkaufman I see. I guess 1.29.1 was accidentally offered to stable Obsidian users on 1.8.10, which is why my Lint on Save broke. I guess other users in my situation will just have to wait, or figure out to downgrade.

pjkaufman:
Yeah. That is correct. I missed the version not being right in `versions.json` which allowed users to be able to download the version 1.29.1 when they should have not been able to.

redactedscribe:
Sorry for all the questions, I was just confused. I likely still don't understand it all. I just hope other 1.8.10 users will be upgraded from 1.29.1 in the near future and won't be stuck on that version forever. Thank you for your help @pjkaufman.

PR #1378, opened 2025-09-02
Add a 'Suppress message if no change' feature with settings.

Moyf:
It's the new version with settings of #1286 

I'm sorry I forgot about this earlier... Since there were some issues with the original PR (my merge operation caused a mess), I'm opening a new one now.  

It now has separate settings, disabled by default:

Snipaste_2025-09-02_11-16-52


src/ui/linter-components/tab-components/general-tab.ts
```diff
@@ -54,6 +54,10 @@ export class GeneralTab extends Tab {
       displayLintOnActiveFileChangeSetting.hide();
     }
 
+    tempDiv = this.contentEl.createDiv();
+    const suppressMessageWhenNoChangeSetting = new ToggleSetting( tempDiv, 'tabs.general.suppress-message-when-no-change.name' as any, 'tabs.general.suppress-message-when-no-change.description' as any, 'suppressMessageWhenNoChange', this.plugin);
```
pjkaufman, line 58: Could you explain the use of `as any` here? The other areas are not using as any. So this implies that it was either an accidental addition or the string provided does not map properly for the English translation.

pjkaufman (changes requested):
Other than passing in the string references `as any` this looks good to me. If you could remove the `as any` then I can merge this. Otherwise I can remove the `as any` if I do not hear back or if that is easier.

src/ui/linter-components/tab-components/general-tab.ts
Moyf: > Could you explain the use of `as any` here? The other areas are not using as any. So this implies that it was either an accidental addition or the string provided does not map properly for the English translation.

Oh, sorry for late replying! I'll try to update this commit, but I'm worried about causing issues with multi-branch operations (just like last time...).  
If I don't succeed, feel free to remove this `as any` :D

pjkaufman (approved):
Looks good. Thanks!

PR #1393, opened 2025-09-19
Update Traditional Chinese Translation for Characters Added and Removed

emisjerry:
update zh-tw.ts for 1.30.0
Use zh-tw.ts for Traditional Chinese


pjkaufman:
Thanks for the translation update. I will go ahead and merge this.

emisjerry:
Hi, 

I've updated to v1.31.0, and zh-tw used, but the language still displayed as zh-cn. Is my setting incorrect?

image

PR #1403, opened 2025-09-24
Fix srcToken error

### LINKED ISSUE #1402: Bug: Horizontal rule creates error with YAML parsing

## Describe the Bug

Consider this note content

```
# Heading A

---

# Heading B
```

When linting this file (e.g. with Yaml Key Sort) it produces an error: `Error "YAML Key Sort" encountered an unknown error: Cannot read properties of null (reading 'srcToken')`

Image

~~Possibly it mistakes the content before `---` as YAML?~~ This seems due to no existing YAML

## How to Reproduce

Lint the file with content above

## Expected Behavior

Linter should act as if this file has no YAML content (which it hasn't). Input should be similar to output (with my applied rules)

## Device

- [x] Desktop (Obsidian 1.9.13)
- [ ] Mobile

mdbraber:
Fixes #1402 @pjkaufman could you review?

### DIFF
--- a/src/rules/yaml-key-sort.ts
+++ b/src/rules/yaml-key-sort.ts
@@ -63,9 +63,9 @@ export default class YamlKeySort extends RuleBuilder<YamlKeySortOptions> {
     const doc = parseYAML(yamlText);
     const startingPriorityKeys = getEmptyDocument(doc);
 
-	if(doc.contents == null) {
-		return text;
-	}
+    if (doc.contents == null) {
+      return text;
+    }
 
     let remainingKeys = this.getYAMLKeysSorted(yamlKeys, doc, startingPriorityKeys);
 


pjkaufman:
Hey @mdbraber , thanks for creating a potential fix for the issue. My concern right now is how it got as far as it did in the logic since #1402 implies that the logic actually ran for yaml key sort. The logic should not make it past this check: https://github.com/platers/obsidian-linter/blob/cce4f98818215d8d3d8d368210368ff620d08d32/src/rules/yaml-key-sort.ts#L41-L44

If there is no YAML, then it should just skip the rule as a whole. So I am more worried that this is a bandaid that hides the reason that `getYAMLText` is returning a non-null value for a file without YAML.

pjkaufman:
It has been a little a while on this. I am going to just go ahead and merge this since it looks to fix the issue.

mdbraber:
Thanks Peter!

PR #1423, opened 2025-11-06
fix: Add missing micromark-extension-frontmatter dependency

MarkShawn2020:
## Summary
- Fixed build errors and runtime module resolution issues
- Added missing `micromark-extension-frontmatter` dependency
- Properly configured esbuild to bundle third-party dependencies

## Problem
The plugin failed to load in Obsidian with error:
```
Cannot find module 'micromark-extension-frontmatter'
```

## Root Cause
- `micromark-extension-frontmatter` was imported in `src/utils/mdast.ts` but not listed in package.json dependencies
- Obsidian plugins must bundle all third-party dependencies (only `obsidian` and Node.js built-ins should be external)

## Solution
1. Added `micromark-extension-frontmatter` to package.json dependencies
2. Ensured it's properly bundled by esbuild (not marked as external)
3. Added `additionalExternal` parameter to esbuild config for test-specific external modules (e.g., `expect`)

## Changes
- `package.json`: Added micromark-extension-frontmatter to dependencies
- `esbuild.config.mjs`: Added additionalExternal parameter for flexible external configuration
- Bundle size increased from 741K to 744K (includes frontmatter extension)

## Test plan
- [x] Build succeeds without errors: `pnpm build`
- [x] Plugin loads successfully in Obsidian
- [x] No runtime module resolution errors
- [x] YAML frontmatter parsing works correctly

### DIFF
--- a/esbuild.config.mjs
+++ b/esbuild.config.mjs
@@ -65,7 +65,7 @@ const unusedCodeForProduction = [replace({
   delimiters: ['', ''],
 })];
 
-const createEsbuildArgs = function(banner, entryPoint, outfile, extraPlugins) {
+const createEsbuildArgs = function(banner, entryPoint, outfile, extraPlugins, additionalExternal = []) {
   return {
     banner: {
       js: banner,
@@ -78,7 +78,8 @@ const createEsbuildArgs = function(banner, entryPoint, outfile, extraPlugins) {
     bundle: true,
     external: [
       'obsidian',
-      ...builtins],
+      ...builtins,
+      ...additionalExternal],
     format: 'cjs',
     target: 'es2020',
     sourcemap: prod ? false : 'inline',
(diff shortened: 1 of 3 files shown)


pjkaufman:
Hey @MarkShawn2020 , could you elaborate on how to reproduce the error in question? I have yet to receive an actual bug report on this and would like to make sure I am actually testing the right version of logic when I look at things tomorrow.

Also, could you elaborate on why you added the pnpm lock file? That is not really the package manager used by this plugin, so adding it does not make sense to me unless there is a very clear benefit to doing so.

Also, why is expect added as an external package in the integration tests? I am not aware of any reason to do that as the code currently works for those.

pjkaufman:
I am not able to reproduce the error in question when the plugin loads. So if this is an issue, please let me know. If I don't hear back by Monday, I will just go ahead and close this PR.

pjkaufman:
I am closing this since I have not heard back as to how to reproduce the issue.

PR #1427, opened 2025-11-14
fix: escape markdown characters

### LINKED ISSUE #1426: Bug: File Name Heading doesn't escape markdown special characters
- [X] I have verified that I am on the latest version of the Linter

## Describe the Bug

`File Name Heading` rule doesn't escape markdown special characters.

## How to Reproduce

1. Create empty note `Escape [_].md`
2. Run `File Name Heading` rule.
3. Observe created header `# Escape[_]`

## Expected Behavior

``` markdown
# Escape \[\_\]
```

## Device

- [X] Desktop
- [X] Mobile

mnaoumov:
Fixes #1426

### DIFF
--- a/__tests__/yaml-title.test.ts
+++ b/__tests__/yaml-title.test.ts
@@ -374,5 +374,17 @@ ruleTest({
         # Hello world
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1428
+      testName: 'Unescapes H1 markdown special characters',
+      before: dedent`
+        # Escape \\[\\_\\]
+      `,
+      after: dedent`
+        ---
+        title: Escape [_]
+        ---
+        # Escape \\[\\_\\]
+      `,
+    },
   ],
 });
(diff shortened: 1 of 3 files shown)


mnaoumov:
I added the missing `jsdoc` but I wonder why do you use `jsdoc` instead of `tsdoc`.

pjkaufman:
> I added the missing `jsdoc` but I wonder why do you use `jsdoc` instead of `tsdoc`.

This is the first I have heard of `tsdoc`. So that is probably why I use `jsdoc`.

pjkaufman:
Thanks for taking the time to create this PR, I will take a look at it later this week as I have time.

mnaoumov:
> > I added the missing `jsdoc` but I wonder why do you use `jsdoc` instead of `tsdoc`.
> 
> This is the first I have heard of `tsdoc`. So that is probably why I use `jsdoc`.

`jsdoc` doesn't really make sense in `TypeScript` project. [`tsdoc`](https://tsdoc.org/) is the way to go. `jsdoc`, for example, duplicates parameter types, which makes sense only in `JavaScript`, but not in `TypeScript`

mnaoumov:
I should add another test case where the rendered version differs

PR #1431, opened 2025-11-23
test: better example

mnaoumov:
Better example showing why #1427 is necessary


pjkaufman (approved):

pjkaufman:
Thanks for the example. I will go ahead and merge this.

PR #1433, opened 2025-11-23
feat: add rule to remove horizontal rules outside YAML frontmatter

Lymnah:
Implemented a new `RemoveHorizontalRulesOutsideYaml` rule to clean up standalone horizontal rule separators (`---`) that are frequently inserted by ChatGPT and other LLMs into generated content. These separators often clutter the body of notes and can break formatting, so this rule removes them safely while preserving YAML frontmatter boundaries.

image

**Before applying lint:** 
image

**After applyling lint:**
image

Key additions:
- Added `RemoveHorizontalRulesOutsideYaml` rule, which strips standalone `---` lines only in the document body and never inside YAML frontmatter.
- Added `splitYamlAndBody` utility to reliably separate YAML from the rest of the document.
- Added comprehensive test coverage, including:
  - YAML preservation
  - code block protection
  - table row handling
  - mixed-content edge cases
- Added localization entries for English, Russian, Simplified Chinese, and Traditional Chinese.

This feature provides a clean way to sanitize unwanted separators generated by LLMs without affecting valid Markdown or YAML structure.

### 1. YAML frontmatter
- YAML `---` fences are preserved exactly.
- Any `---` appearing **after** the YAML frontmatter is removed.
- Works when the YAML body is empty.
- Works when only YAML exists in the file.
- Handles multiple `---` separators in the body while keeping the YAML block untouched.

### 2. Files without YAML
- Removes all standalone `---` lines when no YAML frontmatter is present.
- Removes a lone `---` when it's the only content in the file.

### 3. `---` that must NOT be removed
- `---` inside Markdown **tables** is preserved.
- `---` inside **code blocks** is preserved.
- Inline `---` (e.g., `text --- more text`) is preserved.

### 4. Edge cases
- Handles `---` with leading/trailing whitespace.
- Handles empty files safely.


mnaoumov:
Just removing `---` might be not the best option. Maybe you should consider transforming it into equivalent form, which doesn't get conflicted with frontmatter.

```markdown
***
___
* * *
- - -
_ _ _
```

pjkaufman:
@Lymnah , thanks for putting in a PR. I would suggest trying to target the hr AST type rather than using regex to specifically target `---`. The change in question might work for very specific cases, but hr elements are not required to be `---` exactly and it can be used as a part of a table or even a header.

pjkaufman:
Also, I am thinking this would be better as a paste rule @Lymnah . Is there a reason it is not a paste rule? Are you having the AI generate the files in the actual vault? If so, I could see it being used as a regular rule.

Just as a note, I have not seen any feedback/response since the last message back in December. I will keep this open a couple more weeks to see if there is any feedback. But if there is no back and forth by then, I will go ahead and close this PR as I have limited time right now and cannot manage trying to get a hold of people while making any changes to the repo.

pjkaufman:
I am going to close this as I have not heard back on my questions around the PR. We can revisit this again later if need be.

PR #1443, opened 2025-12-31
feat: add heading-filename-sync rule for H1/filename synchronization

### LINKED ISSUE #192: Keep first H1 heading in sync with filename
I use the option to insert an H1 heading with the filename. But if I later change the filename the H1 heading no longer reflects the filename.

Could you add an option to keep the two in sync?

Possible complications...
- Should the sync be 2 way? If I change H1 should the filename change?

Existing solutions...
- There is an [existing plugin](https://github.com/dvcrn/obsidian-filename-heading-sync) which offers this function, but it would make sense to include keep all linting type operations in a single plugin.

Thanks for your consideration.

### LINKED ISSUE #1418: FR: 3-way-sync (H1, Title, Filename) with regex prefix handling

Yes. Many users, especially those using Zettelkasten or timestamp-based notes, use a prefix in their filenames (e.g., `YYYYMMDDhhmm_`). The problem is that Linter's current title-related rules apply to the *entire* filename. This means if I want to sync my filename with my H1 or frontmatter `title`, the ugly prefix (e.g., `202510291209_`) gets synced as well. I'm always frustrated that I cannot have a clean H1/frontmatter title that stays in sync with the *semantic part* of my filename.

I would like a new rule (or an enhancement to the existing "YAML Title" rule) that enables **3-way synchronization** between the H1 heading, the `title` key in the frontmatter, and the filename, *while ignoring a specified regex prefix in the filename*.

This rule would need a new setting: **"Regex prefix to ignore in filename"**.

The sync should be multi-directional:

1.  **If I change the filename:** The plugin should strip the prefix, and update the H1 and frontmatter `title` with the *rest* of the name.
2.  **If I change the H1:** The plugin should update the frontmatter `title` *and* the filename. When updating the filename, it must **preserve the original prefix**.
3.  **If I change the frontmatter `title`:** The plugin should update the H1 *and* the filename. When updating the filename, it must **preserve the original prefix**.

**My Regex Setting:** `^\d{12}_`

**Scenario: User changes the H1 Heading**

(File is currently named `202510291209_My Old Note.md`)

```markdown
---
title: My Old Note
---

# My New Note

This is my note.
```

**Expected Linter outcome:**

1.  Linter detects the H1 change.
2.  It updates the frontmatter `title` to "My New Note".
3.  It updates the filename, preserving the prefix, to `202510291209_My New Note.md`.

**Result:**

```markdown
---
title: My New Note
---

# My New Note

This is my note.
```

(And the file is renamed)

I have considered the `File Title Updater` plugin, but it does not support ignoring prefixes; it syncs the *entire* filename. I have also tried the existing Linter rules (like "YAML Title" with the "Filename" mode), but they are one-way (Filename -\> Title) and do not handle the multi-directional sync I need (e.g., H1 -\> Filename).

## Additional Context

This feature would be incredibly valuable for anyone using a Zettelkasten (timestamp-based) naming convention. It bridges the gap between having a machine-readable, unique filename (for sorting and ID) and a human-readable, clean title (for reading and linking).

tjmgregory:
## Summary

Implements the long-requested feature to keep H1 headings and filenames synchronized:

- **New rule: `heading-filename-sync`** - Keeps the first H1 heading synchronized with the filename
- **Three sync directions**:
  - `heading-to-filename` (default): Rename the file to match the H1 heading
  - `filename-to-heading`: Update the H1 heading to match the filename  
  - `bidirectional`: Keep both in sync (heading wins on conflict)
- **Regex prefix/suffix patterns** for Zettelkasten and similar workflows (e.g., strip `^\d{12}_` timestamps)
- **Conflict detection** with the existing `file-name-heading` rule
- **User notification** on rename failures
- **Ignore files** using regex

## Related Issues

Closes #192 - FR: Keep first H1 heading in sync with filename (opened 2022)
Closes #1418 - FR: 3-way-sync (H1, Title, Filename) with regex prefix handling

## Changes

- New rule: `src/rules/heading-filename-sync.ts`
- Tests: `__tests__/heading-filename-sync.test.ts` (36 test cases)
- Translations: Updated `src/lang/locale/en.ts`
- Rules runner: Added `PendingRename` mechanism for file rename signaling
- Main plugin: Added file rename handling via `app.fileManager.renameFile()`
- Conflict detection between `heading-filename-sync` and `file-name-heading`

### File Renaming Architecture

Rules can only return modified text, so file renaming uses a callback pattern:

1. Rule signals rename via `setPendingRename` callback passed in options
2. `RulesRunner` stores the pending rename in `this.pendingRename`
3. After `lintText()` returns, `main.ts` checks for pending rename and executes it

### Sync Direction Behavior

| Direction | Behavior |
|-----------|----------|
| `heading-to-filename` | Renames file to match H1 (default, preserves prefix/suffix) |
| `filename-to-heading` | Updates H1 to match filename |
| `bidirectional` | No H1? Insert from filename. H1 differs? Rename file to match. |

### Integration with Existing Rules

- **`yaml-title`**: Enable both for 3-way sync (H1, filename, YAML title)
- **`file-name-heading`**: Conflict detection added - users prompted to choose one

## Test Plan

- [x] All 1211 tests pass (including 36 new tests)
- [x] Build passes (`npm run build`)
- [x] Linting passes (`npm run lint`)


tjmgregory:
## Bug Fix: Race Condition in Folder Linting (commit d71dc1e)

Fixed a critical bug where "Lint folder" operations would cause incorrect file renaming.

### The Problem
When linting an entire folder, files were processed in parallel using `Promise.all`. The `pendingRename` was stored as a shared instance property on `RulesRunner`, causing one file's rename to overwrite another's.

This resulted in:
- File titles not being updated correctly
- One file's heading accidentally renaming a different file

### The Solution
Changed `lintText()` to return a `LintResult` object containing both the linted text and the `pendingRename`. Each parallel operation now gets its own isolated result, eliminating the race condition.

### Changes
- Added `LintResult` type: `{text: string, pendingRename: PendingRename | null}`
- Removed `pendingRename` instance property from `RulesRunner`
- Updated `lintText()` to return `LintResult` instead of `string`
- Updated `handlePendingRename()` to accept `pendingRename` as parameter
- Updated `runLinterFile()` and `runLinterEditor()` to use `LintResult`

tjmgregory:
## Bug Fix: Case-Only Renames on macOS/Windows (commit 1e5db79)

Fixed an issue where renaming files with only case changes would fail on case-insensitive filesystems.

### The Problem
On macOS and Windows (case-insensitive by default), renaming a file where only the case changes fails:
- `Rules for showing the checklist.md` → `Rules for Showing the Checklist.md`

The OS sees these as the "same" file and returns "file already exists".

### The Solution
Detect case-only renames by comparing lowercase paths. When detected, use a two-step rename via a temporary intermediate name:

```
1. file.md → file.md.linter-temp-rename
2. file.md.linter-temp-rename → File.md
```

This workaround is commonly used to handle case-insensitive filesystem limitations.

tjmgregory:
@pjkaufman How can I help to get this reviewed/merged? 🙂

pjkaufman:
My current concern is that this looks to be AI generated. I need to take a closer look at the code, but how is it handling changes from outside Obsidian? For example, what happens when someone changes the file name, YAML title, and H1 all to be different values outside of Obsidian or while the Linter is disabled. When the Linter runs against the file, which of those 3 values will be used for the heading, YAML title, and file name?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+    super({
+      nameKey: 'rules.heading-filename-sync.name',
+      descriptionKey: 'rules.heading-filename-sync.description',
+      type: RuleType.HEADING,
+      ruleIgnoreTypes: [IgnoreTypes.code, IgnoreTypes.math, IgnoreTypes.yaml, IgnoreTypes.link, IgnoreTypes.wikiLink, IgnoreTypes.tag],
+      hasSpecialExecutionOrder: true,
+      disableConflictingOptions(value: boolean, app: App): void {
+        // Check for conflict with file-name-heading rule
+        const fileNameHeadingOptions = rulesDict['file-name-heading'];
+        if (!fileNameHeadingOptions) return;
```
pjkaufman: I see that we are checking whether this is empty or not. However I am not sure how this would happen given we fill out missing rules when the plugin loads and the app is in the "onlayoutready" stage.

What does this check actually do beyond exit early for a situation we should not encounter? Or is this scenario actually happening?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+      prefixPattern: string,
+      suffixPattern: string,
+  ): string {
+    let result = fileName;
+
+    if (prefixPattern) {
+      try {
+        const prefixRegex = new RegExp(prefixPattern);
+        result = result.replace(prefixRegex, '');
+      } catch (e) {
```
pjkaufman, line 97: I am probably missing something here, but what about this is guarantying that the prefix regex only affects the start of the file name?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+        // Invalid regex, skip prefix stripping
```
pjkaufman: If the regex is invalid, we should at the very least console log it so the user knows that the prefix regex is invalid.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+      }
+    }
+
+    if (suffixPattern) {
+      try {
+        const suffixRegex = new RegExp(suffixPattern + '$');
+        result = result.replace(suffixRegex, '');
+      } catch (e) {
+        // Invalid regex, skip suffix stripping
```
pjkaufman: If the regex is invalid, we should at least log that it is invalid so the user can at least know that they need to try to fix it.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+  private extractFilenamePrefix(
+      fileName: string,
+      prefixPattern: string,
+  ): string {
+    if (!prefixPattern) {
+      return '';
+    }
+
+    try {
+      const prefixRegex = new RegExp(prefixPattern);
```
pjkaufman: We should create the regexp once and reuse it across the functions. Besides, if we already pulled the regex prior, why are we once again running the prefix regex again? Can't we grab the value in one step?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+    if (!suffixPattern) {
+      return '';
+    }
+
+    try {
+      const suffixRegex = new RegExp(suffixPattern + '$');
+      const match = fileName.match(suffixRegex);
+      return match ? match[0] : '';
+    } catch (e) {
+      return '';
```
pjkaufman: Same issue as with the prefix regex.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+  get OptionsClass(): new () => HeadingFilenameSyncOptions {
+    return HeadingFilenameSyncOptions;
+  }
+
+  apply(text: string, options: HeadingFilenameSyncOptions): string {
+    const semanticFilename = this.extractSemanticFilename(
+        options.fileName,
+        options.filenamePrefix,
+        options.filenameSuffix,
+    );
```
pjkaufman: I am not sure why you cannot get all 3 name parts right here if you are just going to run the prefix and suffix regex.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+
+    const existingH1Match = text.match(/^#\s+(.*)$/m);
```
pjkaufman, line 63: Rather than duplicate the regex, if we are going to do this, we will want to extract the regex from https://github.com/pjkaufman/obsidian-linter/blob/7c123c37f23b95ee531545951acf31e0c5a6ee21/src/utils/regex.ts#L119 and make it into a global regex. Then we can reuse it there and import it here and use it.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+  private syncFilenameToHeading(
+      text: string,
+      semanticFilename: string,
+      existingH1Match: RegExpMatchArray | null,
+  ): string {
+    const escapedFilename = escapeMarkdownSpecialCharacters(semanticFilename);
+    const newHeading = `# ${escapedFilename}`;
+
+    if (existingH1Match) {
+      return text.replace(/^#\s+.*$/m, newHeading);
```
pjkaufman, line 127: Doesn't the regexp match array already let you know what the location of the text you need to replace is? If so, why are we doing another regex find and replace?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+    // If the heading matches the semantic filename, no rename needed
+    if (existingH1Unescaped === semanticFilename) {
+      return text;
+    }
+
+    // Build the new filename with preserved prefix/suffix
+    const prefix = this.extractFilenamePrefix(options.fileName, options.filenamePrefix);
+    const suffix = this.extractFilenameSuffix(options.fileName, options.filenameSuffix);
+    const sanitizedHeading = this.sanitizeFilename(existingH1Unescaped);
+    const newBasename = `${prefix}${sanitizedHeading}${suffix}`;
```
pjkaufman: Why would the prefix and suffix be preserved if it is not present in the heading and the heading has changed? As someone who does not fully understand the use case, this seems odd to me. I can rationalize doing this if we assume that the prefix and suffix will never change. But I find that hard to believe that there will not be a scenario where the user doesn't want to retain the prefix or suffix.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+  ): string {
+    // If there's no H1, sync filename to heading (insert new heading)
+    if (!existingH1Unescaped) {
+      return this.syncFilenameToHeading(text, semanticFilename, existingH1Match);
+    }
+
+    // If H1 differs from filename, heading wins - rename file
+    if (existingH1Unescaped !== semanticFilename) {
+      return this.syncHeadingToFilename(text, options, existingH1Unescaped, semanticFilename);
+    }
```
pjkaufman, line 197: Is this not something that should be a setting that user decides on?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+  private sanitizeFilename(heading: string): string {
+    // Remove/replace characters that are invalid in filenames
+    // Invalid chars: / \ : * ? " < > |
+    return heading
+        .replace(/[/\\:*?"<>|]/g, '-')
+        .replace(/-+/g, '-') // Collapse consecutive dashes
+        .replace(/^-+|-+$/g, '') // Remove leading/trailing dashes
+        .replace(/\s+/g, ' ')
+        .trim();
+  }
```
pjkaufman, line 212: Is this fine on all the major OSes? I am not sure of the exact list, but this seems like it might be tailored to a specific OS just based on looking at it.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+    return text;
+  }
+
+  private sanitizeFilename(heading: string): string {
+    // Remove/replace characters that are invalid in filenames
+    // Invalid chars: / \ : * ? " < > |
+    return heading
+        .replace(/[/\\:*?"<>|]/g, '-')
+        .replace(/-+/g, '-') // Collapse consecutive dashes
+        .replace(/^-+|-+$/g, '') // Remove leading/trailing dashes
```
pjkaufman, line 209: These two are just stylistic and not actual sanitization steps to make sure that it is valid. Is there a reason for this?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+        .replace(/\s+/g, ' ')
```
pjkaufman, line 210: Same issue with this line. It seems like just cleaning up the file name instead of using what has been provided. It is not what I would expect to be in this function or what I would expect given the intent of the logic which is to use the H1 as the file name or vice versa.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+        continue;
+      }
+
+      try {
+        const regex = new RegExp(pattern);
+        if (regex.test(filePath)) {
+          return true;
+        }
+      } catch (e) {
+        // Invalid regex, skip this pattern
```
pjkaufman: Invalid regex should result in a console log at the very least.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+        `,
+        after: dedent`
+          # Meeting Notes
+          Content here.
+        `,
+        options: {
+          fileName: '202312151030_Meeting Notes',
+          filenamePrefix: '^\\d{12}_',
+          syncDirection: 'filename-to-heading',
+        },
```
pjkaufman, line 259: Note: this functionality could be added to the existing rule instead of as a new rule given it only seems to be on this rule because it was meant to be a 2 way sync.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+      return text.replace(/^#\s+.*$/m, newHeading);
+    }
+
+    let yaml_end = text.indexOf('\n---');
+    yaml_end = yaml_end === -1 || !text.startsWith('---\n') ? 0 : yaml_end + 5;
+
+    let header = `${newHeading}\n`;
+    if (text.length < yaml_end) {
+      header = '\n' + header;
+    }
```
pjkaufman, line 136: If I am understanding correctly, this just adds the new heading on its own line.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+            value: 'heading-to-filename',
+            description: 'Rename the file to match the H1 heading',
+          },
+          {
+            value: 'filename-to-heading',
+            description: 'Update the H1 heading to match the filename',
+          },
+          {
+            value: 'bidirectional',
+            description: 'Keep heading and filename in sync (heading wins on conflict)',
```
pjkaufman, line 301: Personally, I think that the user should decide which wins in the case of a conflict.

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+        nameKey: 'rules.heading-filename-sync.filename-suffix.name',
+        descriptionKey: 'rules.heading-filename-sync.filename-suffix.description',
+        optionsKey: 'filenameSuffix',
+      }),
+      new TextAreaOptionBuilder({
+        OptionsClass: HeadingFilenameSyncOptions,
+        nameKey: 'rules.heading-filename-sync.ignore-rename-files.name',
+        descriptionKey: 'rules.heading-filename-sync.ignore-rename-files.description',
+        optionsKey: 'ignoreRenameFiles',
+      }),
```
pjkaufman: Could you explain what the point of this is? We can already ignore files if we don't want any rules enabled. If we don't want a rule to run, we have the ability to specify it in the frontmatter. Is this just meant to be a convenience thing to avoid having to specify in the frontmatter which files should have this rule disabled?

If so, is there a better way to handle this that would benefit the other rules as well instead of just making a one off solution for this one rule?

src/rules/heading-filename-sync.ts
```diff
@@ -0,0 +1,383 @@
...
+
+    let yaml_end = text.indexOf('\n---');
+    yaml_end = yaml_end === -1 || !text.startsWith('---\n') ? 0 : yaml_end + 5;
+
+    let header = `${newHeading}\n`;
+    if (text.length < yaml_end) {
+      header = '\n' + header;
+    }
+
+    return insert(text, yaml_end, header);
```
pjkaufman, line 138: This may be a nitpick, but I don't see a point in using insert in the cases where you are just adding text to the end of the file's contents.

pjkaufman (changes requested):
This is an initial review of just the rule and not the other things it affects. I have yet to be able to review most of the changes outside the new rule itself. So that still needs to be looked at. Some of this looks fine, but other parts make me question whether to consider this out of scope for the Linter since touching the file name has traditionally been considered out of scope for the Linter. It can read it, but changing it opens a whole can of worms that I am not currently available to tackle.

pjkaufman:
I am planning on closing this PR if I don't hear back by the end of this next week.

tjmgregory:
> I am planning on closing this PR if I don't hear back by the end of this next week.

Hey there @pjkaufman - I apologise for the delay on this. Lost in a slew of work repo notifications! I'm looking at this today and will get some replies sent through to your other comments. Thanks for the feedback 🙏

tjmgregory:
> My current concern is that this looks to be AI generated. I need to take a closer look at the code, but how is it handling changes from outside Obsidian? For example, what happens when someone changes the file name, YAML title, and H1 all to be different values outside of Obsidian or while the Linter is disabled. When the Linter runs against the file, which of those 3 values will be used for the heading, YAML title, and file name?

The intention is that whichever "sync direction" you've chosen becomes the winner when the linter next runs. I'd not weighed in the yaml title in this work, but would be happy to if you felt it was a consistent enough field to be merited 👍

To be transparent I do use AI to write my code these days, but I'm intentional when raising external reviews and do vet code contributions to others' repos to a level I'm happy with. Appreciate that this surge in _capacity_ to code has no doubt put a large burden on all popular open source maintainers with _quality_-yearning PRs appearing all over the shop! 😅

src/rules/heading-filename-sync.ts
tjmgregory: Roger that - was just a lack of insight into it's population and so played it safe with an index access - will remove 👍

src/rules/heading-filename-sync.ts
tjmgregory: I'd not thought about that - good call. Will wrap each in some start/end capture patterns 👍

src/rules/heading-filename-sync.ts
tjmgregory: Yep, much cleaner will do 👌

src/rules/heading-filename-sync.ts
tjmgregory: So, this came from using Zettelkasten system at the time and know that for those users you capture metadata in your filenames directly that you wouldn't necessarily want bleeding into your title - and so this helps to defend against that. Happy to debate if this is truly warranted though or whether you'd want to just add it later should people request it?

src/rules/heading-filename-sync.ts
tjmgregory: So I'd avoided that in the first pass, just because it felt like loading up too many options and this use-case is minor. But actaully having sat on this for a little while, I'm not sure _quite_ how useful this is. I'd be open to just nothing happening if the "source" locaiton wasn't present, which would only be the case for the heading (and frontmatter if support added)

src/rules/heading-filename-sync.ts
tjmgregory: Will confirm

src/rules/heading-filename-sync.ts
tjmgregory: Sure, this immediate use case came from my Daliy Notes setup, where I have `01 Daily Notes/<year>/<month>/<day>.md`. These get templated with `# Daily Notes - YYYY-MM-DD` and I didn't want this happening for this whole directory and so this provided the escape hatch.

Do see that this does make for a complex config for a single rule - and also see that this would be something others may benefit from if designed right. Conscious getting this right could be a whole can of worms, but let me take a pass at a couple of ideas just to at least open the discussion and see if anything happens to feel right enough.

src/rules/heading-filename-sync.ts
tjmgregory: It's not necessarily just adding to the end, if there's already text in teh docuemnt you want it to remain and to slot the title inbetween the front matter and the rest of what's in the doc 🙂

tjmgregory:
> This is an initial review of just the rule and not the other things it affects. I have yet to be able to review most of the changes outside the new rule itself. So that still needs to be looked at. Some of this looks fine, but other parts make me question whether to consider this out of scope for the Linter since touching the file name has traditionally been considered out of scope for the Linter. It can read it, but changing it opens a whole can of worms that I am not currently available to tackle.

Could you elaborate on what some of the worms are? 🪱 Hadn't considered this - certainly don't want this to have to lead to a whole load of other maintenance

src/rules/heading-filename-sync.ts
tjmgregory: I guess in hindsight, I could have just put this ignore rule in the template... maybe that's all that's needed. I'll strip this and confirm that actually solves all my problems 👍

src/rules/heading-filename-sync.ts
tjmgregory: Yeh so I flipped and flopped on this, but very happy to take your lead. My rationale was possibly a little shortsighted and mainly just not wanting disruprt other people's flows. We could relatively easily just replace it with this new rule and the file name defaulting as the source of truth

pjkaufman:
Sorry about the delay in getting you feedback. I have just gotten back into making a lot of changes for the Linter and decided to pop back in here.

One thing to note is that the file name updates are not something I would like to support in the Linter itself. I am fine with other file content changes, but the file name updates are a show stopper for me on this.

pjkaufman:
> > This is an initial review of just the rule and not the other things it affects. I have yet to be able to review most of the changes outside the new rule itself. So that still needs to be looked at. Some of this looks fine, but other parts make me question whether to consider this out of scope for the Linter since touching the file name has traditionally been considered out of scope for the Linter. It can read it, but changing it opens a whole can of worms that I am not currently available to tackle.
> 
> Could you elaborate on what some of the worms are? 🪱 Hadn't considered this - certainly don't want this to have to lead to a whole load of other maintenance

Most of the can of worms is needing to handle file renames, update anything that is holding onto information about the file name (the logic for which files are being linted, which files have been updated that have to have their content changed, and things like that). Then you have to deal with the possibility that the file name is synced across different OSs. So what is fine on desktop may not be fine on mobile and vice versa. So headers would need to be sanitized.

Issue #883, opened 2023-09-09
Bug: Do not work after updating


Asthestarsfalll:
- [x] I have verified that I am on the latest version of the Linter

## Describe the Bug

After I update linter, It stops working.

And there's no log when I run linter, So I suspect  linter is not running at all.

## How to Reproduce

I am not sure how to reproduce this.

## Screenshots
I try to manually run linter, but nothing changed
![2023-09-09_09-57](https://github.com/platers/obsidian-linter/assets/72954905/222f6c85-3789-4bb4-a64d-d7db3ba50a04)

![2023-09-09_09-53](https://github.com/platers/obsidian-linter/assets/72954905/896d4903-2582-4f86-a971-b8a3d9bf72ff)

## Device

- [x] Desktop: obisidian 1.3.7 on manjaro linux
- [ ] Mobile

pjkaufman:
@Asthestarsfalll , thanks for reaching out.

A couple of questions to better help me try to reproduce the issue:
- Could you go ahead and provide your `data.json` and a sample file to try to reproduce with?
- If you open up developer tools (ctrl+shift+i) to see if there are any errors when you try to run the Linter, is there any output there?
- How are you trying to run the Linter? Logs are only displayed in that area if you run lint the current file

Asthestarsfalll:
@pjkaufman Thank you for your reply.
My `data.json` is:
```json
{
  "ruleConfigs": {
    "escape-yaml-special-characters": {
      "enabled": false,
      "try-to-escape-single-line-arrays": false
    },
    "force-yaml-escape": {
      "enabled": false,
      "force-yaml-escape-keys": ""
    },
    "format-tags-in-yaml": {
      "enabled": false
    },
    "format-yaml-array": {
      "enabled": false,
      "alias-key": true,
      "tag-key": true,
      "default-array-style": "single-line",
      "default-array-keys": true,
      "force-single-line-array-style": "",
      "force-multi-line-array-style": ""
    },
    "insert-yaml-attributes": {
      "enabled": false,
      "text-to-insert": "aliases: \ntags: "
    },
    "move-tags-to-yaml": {
      "enabled": false,
      "how-to-handle-existing-tags": "Nothing",
      "tags-to-ignore": ""
    },
    "remove-yaml-keys": {
      "enabled": false,
      "yaml-keys-to-remove": ""
    },
    "yaml-key-sort": {
      "enabled": false,
      "yaml-key-priority-sort-order": "",
[...]

PR #1465, opened 2026-02-23
feat: Support Additional File Extensions

sec-ml:
Added a config option (array) to support additional markdown file formats after finding an old issue (but relevant issue): https://github.com/platers/obsidian-linter/issues/883#issuecomment-3941972360

Obsidian still needs to be able to recognise the other files as markdown (i.e. through the use of other plugins).

I've tested with:

- [obsidian://show-plugin?id=obsidian-markdown-file-suffix](obsidian://show-plugin?id=obsidian-markdown-file-suffix) - works for .mdx, .svx
- obsidian://show-plugin?id=edit-mdx - works for .mdx
- obsidian://show-plugin?id=mdx-as-md-obsidian - works for .mdx
- obsidian://show-plugin?id=qmd-as-md-obsidian - works for .qmd

I've used the same config layout as `folder-ignore-option.ts` as it seemed most appropriate.

I setup a local environment for testing, and saw 6 test failures before making changes, and I see the same failures after.

### DIFF
--- a/src/lang/locale/de.ts
+++ b/src/lang/locale/de.ts
@@ -149,7 +149,7 @@ export default {
       },
       'additional-file-extensions': {
         'name': 'Zusätzliche Dateierweiterungen',
-        'description': 'Dateierweiterungen, die zusätzlich zu md gelinted werden sollen. Zum Beispiel mdx oder svx. Führenden Punkt nicht angeben. Hinweis: Es werden nur Dateien gelinted, die Obsidian als Markdown ansieht (nativ oder über andere Plugins), unabhängig von hinzugefügten Erweiterungen.',
+        'description': 'Dateierweiterungen, die zusätzlich zu md gelinted werden sollen. Zum Beispiel mdx oder svx. Führenden Punkt nicht angeben. <b>Hinweis: Es werden nur Dateien gelinted, die Obsidian als Markdown ansieht (nativ oder über andere Plugins), unabhängig von hinzugefügten Erweiterungen.</b>',
         'extension-placeholder': 'z.B. mdx',
         'add-input-button-text': 'Weitere Erweiterung hinzufügen',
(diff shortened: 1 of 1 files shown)


sec-ml:
For translations: I've added them where there are existing locales. They're AI-generated, but translate back using other tools reasonably well. Not sure if you'd normally wait for other people to manually translate.

pjkaufman:
Thanks for taking the time to create the PR. You beat me to trying to tackle the feature. Right now I am taking a look at the contents to see if anything stands out as a potential issue. I will hopefully have some feedback from the code review soon.

pjkaufman (changes requested):
The changes look pretty good. It looks similar to what I was thinking of adding after the latest comment about wanting mdx and other extensions to work with the Linter. I really just have the suggestion of adding the bold HTML tag around the note in the translations as that is meant to allow for the notes to stand out. But beyond that, this looks pretty good.

I did check the translations just to make sure they were not off base or anything like that. I did not see any glaring issues. The English and Spanish seem good (I cannot really attest to the nativeness of the others). But if users feel that there is a better translation to use, they can always reach out and open an issue or contribute. So I think the translations should be fine. So once the bold is added around the notes, I think this should be good to merge.

sec-ml:
Thanks! Changes made.

I also realised after submitting that there's a project for feature requests - I'll bear that in mind for future.

pjkaufman:
No worries. The projects is mostly meant to help me keep in mind where things are and what I am prioritizing. But I am a bit sporadic and all over the place with that, so don't feel like the only things that can be worked on are things on the project. I just prefer to discuss something before there is a PR for it. Especially when it comes to features that will require more effort in making it continue to work long term since most people who contribute just add what they would like and then disappear (which is fine, but that is to be taken into consideration for maintenance burden in the long term).

PR #1475, opened 2026-03-03
Add 'Do not show again' checkbox to lint confirmation modals

### LINKED ISSUE #1474: Bug: Running lint-all-files command doesn't lint all files but shows warning dialog that can't be suppressed

## Describe the Bug

Running lint-all-files command doesn't lint all files but shows warning dialog that can't be suppressed.

## How to Reproduce

Run "Lint all files" command either through Obsidan UI, command palette, CLI or MCP.

## Expected Behavior

Plugin trusts users to issue commands that they actually want to run or have a setting to disable the warning or only show the warning once.

## Screenshots

Image

## Device

- [x] Desktop
- [x] Mobile

## Additional Context

[Add any other context about the problem here.](https://github.com/platers/obsidian-linter/issues/1473)

jancbeck:
## Summary
- Adds a "Do not show this confirmation again" checkbox to the lint confirmation modals (`lint-all-files` and `lint-all-files-in-folder`)
- When checked and submitted, persists the preference to `data.json` so future invocations skip the modal
- Single shared flag for both commands
- No standalone settings toggle — the checkbox on the modal itself is the only UI surface

## Motivation

Closes #1474

The confirmation modals block automation workflows — running these commands via Obsidian CLI, MCP, or other programmatic invocations shows a dialog that can't be dismissed without user interaction.

Rather than a standalone settings toggle, the preference is set directly from the modal via a "Do not show again" checkbox, which is more discoverable and follows common UI patterns.

## Test plan
- [ ] Run "Lint all files" → modal shows checkbox before buttons
- [ ] Check the box + click Submit → next invocation skips modal entirely
- [ ] Same behavior for "Lint all files in folder"
- [ ] Preference persists across plugin reloads
- [ ] Checkbox absent when `saveSuppressPreference` callback not provided
- [ ] Setting no longer appears in General settings tab


pjkaufman:
I would prefer this be a "do not show again" checkbox on both the folder and lint all modals. It would get stored in the `data.json` in order to persist it, but it would not be its own setting.

pjkaufman:
Going through the CLI will need to be handled via a CLI handler or what have you in order to prevent even asking for the UI confirmation. But that will have to come separately.

jancbeck:
@pjkaufman updated PR.

pjkaufman:
Sorry about the delay. I will go ahead and take a look and see how things look and function.

pjkaufman:
My main concern from just looking at the code is that the code uses the same logic to suppress both modals. But that may not be the intent. I would expect each modal to have its own setting for suppressing seeing it. That hopefully would not be too bad to implement. But I will look at the actual way in which it works in the app.

pjkaufman:
~Well, it looks like it may have just been added to linting all files instead of also for linting a folder. I will double check this though.~

Edit: operator error here. I needed to swap to the branch and check. It looks like both modals do indeed use the same setting, so that would need changing. I can see wanting to suppress the modal for folder lints, but not the lint all functionality as well.

pjkaufman:
I can attest to the logic working. So it just needs to use one boolean for each of the actual modals (folder and all files).

pjkaufman (changes requested):
Looks and seems to work good except for one checkbox being used for both modals. If we can get that swapped over to two settings instead of one, this should be good for merging.

jancbeck:
Split the single `suppressLintConfirmationModal` boolean into two separate booleans:
- `suppressLintAllFilesConfirmationModal` — for the lint-all-files modal
- `suppressLintAllFilesInFolderConfirmationModal` — for the lint-folder modal

Each modal's "Do not show again" checkbox now only affects that specific modal. Added migration in `loadSettings()` to carry over the old setting for existing users.

aamikus:
this is a useful change for agentic workflows. let's get this merged

src/main.ts
```diff
@@ -130,6 +130,14 @@ export default class LinterPlugin extends Plugin {
       this.settings.logLevel = convertNumberToLogLevel(this.settings.logLevel);
     }
 
+    // migrate old single suppress boolean to separate per-modal booleans
+    if (data && typeof data.suppressLintConfirmationModal === 'boolean') {
+      this.settings.suppressLintAllFilesConfirmationModal = data.suppressLintConfirmationModal;
+      this.settings.suppressLintAllFilesInFolderConfirmationModal = data.suppressLintConfirmationModal;
+      delete (this.settings as unknown as Record<string, unknown>).suppressLintConfirmationModal;
+      await this.saveSettings();
+    }
```
pjkaufman: The migration is not necessary here. This is only needed if the change was released. Since it never made a release, it should not be moved.

pjkaufman: I will see about fixing this tomorrow and then double checking that everything still works. If it does, then I will proceed with merging it.

pjkaufman (approved):
Looks good now that I removed the logic around migrating a setting that does not exist.

aamikus:
Now that this is merged, could we make a release with this change?

pjkaufman:
I am still working on a couple of things before the next release. I am hoping to get a couple more changes for the new community store as well as at least one other change in. But we shall see how things go. I might just make a release this weekend if I don't get as far as I wanted on the other things.

pjkaufman:
A new release has been made. I did not end up being able to make time to get the other things in. But the changes should now be out there for this change.

PR #1493, opened 2026-03-22
Clarify nested folder ignoring in documentation

grantwinney:
The way the text is currently phrased, it makes it sound like the "templates" folder will be ignored, but nested folders will be allowed (to be updated by Linter).

It could just be me, but I think that maybe using the phrase "ignored as well" makes it clear that subfolders of the folder will all be ignored. Unless I'm misunderstanding the behavior, and they're actually _not_ ignored?

### DIFF
--- a/docs/docs/usage/disabling-rules.md
+++ b/docs/docs/usage/disabling-rules.md
@@ -11,7 +11,7 @@ The values in the text box are expected to be folder paths from the base of the 
 
 ![Setting for ignoring specific folders](../assets/folders-to-ignore.jpg)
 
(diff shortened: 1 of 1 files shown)


pjkaufman:
Gotcha. I can see how that would be confusing. The line about nested folders is meant to mean that the path provided can be to nested folders as well as folders at the base of the vault. So this clarification works. But I will need to add a line or two to make sure that I note that you can ignore nested folders and their contents without ignoring their parents.

grantwinney:
@pjkaufman Thanks for explaining that! I see what it was trying to convey now.

Issue #1105, opened 2024-06-20
FR: improve YAML key sort to "at Start" and "at End" order


l2eused:
## Is Your Feature Request Related to a Problem? Please Describe.

Yes, I want to sort oder that has prior at top and bottom but plugin doesn't allow it.

## Describe the Solution You'd Like

Add more `YAML Key Priority Sort Order` settings like

```
YAML Key Priority Sort Order (Start): date created, date modified
YAML Key Priority Sort Order (End): tags
YAML Sort Order for Other Keys: asc alphabet
```

also remove `Priority Keys at Start of YAML` since it will use 2 insert box rules (YAML Key Priority Sort Order) and it no need this anymore.

**Please include an example where applicable:**

set `top1` `top2` as `YAML Key Priority Sort Order (Start)` and `bottom1` as `YAML Key Priority Sort Order (End)` and set `YAML Sort Order for Other Keys` to `None`
``` markdown
unsort1
top2
unsort3
bottom1
unsort2
```

``` markdown
top2
unsort1
unsort3
unsort2
bottom1
```

## Additional Context

I want sort order `created` and `updated` (track create/modify time) always on top. **at same time**, I also want to sort `aliases` and `tags` at the bottom. would appreciate if you impliment (improve) this feature!

Graywaren:
Came here to make this feature request myself. I really don't want to have to list every single property i use in between ones i want at the top and bottom.

Nitero:
Would also love this feature.

Alternatively it could stay as one setting, but support regex to give even more control over the middle part.
So your example could look like this:
```
created
updated
.* (any property)
aliases
tags
```
If the order of the middle properties also matters it could look like:
```
created-date
updated-date
.*-date (any date property)
category
type
.* (any property)
aliases
tags
```

pjkaufman:
I would definitely like this to be discussed more prior to an implementation being thrown out there. Right now the sort order is based on what the user specifies for the keys. I am open to the idea of adding two more properties:
- Starting Keys
- Ending Keys

I don't particularly like having to parse regex out of strings since that requires knowing which strings are regex versus which are just plain old key names. 

But I am open to discussion around how this would be implemented. I don't have the time right now to focus on implementing this, but I can discuss it and if @mxberner wants to revise the PR that already exists based on discussions and what is decided, I would be fine with accepting a PR at that point. But I think right now my preference is for two new settings unless there is a compelling reason to use some other method like regex.

PR #1494, opened 2026-03-31
feat: Priority Sort order for unspecified keys (<others> functionality)

mxberner:
## Summary
Addressing #1105 

Allows assigning a "YAML Key Priority Sort Order" for all unspecified keys by including `<others>` as a placeholder in the sort order. This feature enables "sandwiching" undefined keys between a "front" and "back" end of explicitly defined key orderings. 

## PR Known Functionality
- within `<others>` class, "ascending, descending, and none" sorts still function as expected. 
- "Priority Keys at Start of YAML" is functionally redundant — false is achieved through placing `<others>` at top of sort order, true is achieved by placing `<others>` at bottom of sort order. 
- "Priority Keys at Start of YAML" option is overridden iff `<others>` is provided. 
- English translation settings updated to communicate the change.

## Example

Before:
Screenshot 2026-03-31 at 12 56 54 AM

After:
Screenshot 2026-03-31 at 12 57 19 AM

*i.e. `misc1` and `misc2`, not defined in the sort order, can have intelligent placement among the specified fields after linted*

### DIFF
--- a/src/rules/yaml-key-sort.ts
+++ b/src/rules/yaml-key-sort.ts
@@ -8,6 +8,22 @@ import {FlowCollection} from 'yaml/dist/parse/cst';
 
 type YamlSortOrderForOtherKeys = 'None' | 'Ascending Alphabetical' | 'Descending Alphabetical';
 
+const OTHERS_TOKEN = '<others>';
+
+function normalizeYamlSortToken(token: string): string {
+  token = token.trim();
+
+  if (token === OTHERS_TOKEN) {
+    return token;
+  }
+
+  if (token.endsWith(':')) {
+    return token.substring(0, token.length - 1).trimEnd();
+  }
+
+  return token;
+}
+
 class YamlKeySortOptions implements Options {
   priorityKeysAtStartOfYaml?: boolean = true;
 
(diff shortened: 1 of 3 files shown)


src/rules/yaml-key-sort.ts
```diff
@@ -67,27 +74,69 @@ export default class YamlKeySort extends RuleBuilder<YamlKeySortOptions> {
...
+      return this.getTextWithNewYamlFrontmatter(
+          text,
+          oldYaml,
+          astToString(startingPriorityKeys),
+          astToString(doc),
+          priorityAtStartOfYaml,
+          options.dateModifiedKey,
+          options.currentTimeFormatted,
+          options.yamlTimestampDateModifiedEnabled,
+      );
```
pjkaufman, line 123: Maybe I need to read this for more time to better understand what it is trying to do. But I am going to just go ahead and ask. What is the point of making the simple else logic here so much more convoluted?

src/rules/yaml-key-sort.ts
```diff
@@ -8,6 +8,22 @@ import {FlowCollection} from 'yaml/dist/parse/cst';
 
 type YamlSortOrderForOtherKeys = 'None' | 'Ascending Alphabetical' | 'Descending Alphabetical';
 
+const OTHERS_TOKEN = '<others>';
```
pjkaufman, line 11: Note: I am not a fan of placeholders like this. I am not fully aware of what is being attempted by the placeholder other than to specify all other values go there, but I am not sure I like this.

pjkaufman:
I am closing this due to the fact that I have not received any responses to my questions. This can be revisited if need be. I would like the idea of how to do this to be discussed further before an attempt is made to do this.

PR #1500, opened 2026-04-10
Fix: Restore Inline Code Correctly When Moving Footnotes to the Bottom

### LINKED ISSUE #1439: Bug: Footnotes -> Move Footnotes to the Bottom is incorrectly updating/modifying and moving INLINE CODE values
I am on version 1.30.0 which is the latest version!

## Describe the Bug

I enabled `Footnotes -> Move Footnotes to the Bottom` while it properly moves the footnotes, there is a side effect where it is moving content of inline code blocks (anything inside  single ticks example: `CTRL + X`), making it completely unusable. 

It's like all the contents of the inline code blocks are moved 1 place forward. Where the original was, it moves that content to the next time an inline code is used etc. Please see SCREENSHOTS clearly showing the situation.

Note: If I disable this rule, issue immediately goes away.

## How to Reproduce

PLEASE SEE SCREENSHOTS BELOW.

I am attaching my `data.json` for all values I have on the Linter, and the file `Obsidian Syntax`. 

1. Open the OBSIDIAN SYNTAX file and go to 'Meta Bind' Heading
2. See the locations of the inline code blocks like `Meta Bind: Open Playground`
3. Run the linter on the file.
4. See how it modifies the location of the inline code blocks. It seems to be moving the contents around.

Or

The entire file has multiple different places that have inline code, and almost all of the occurrences that are multiword are moved around. This is easily noticeable if you have the git plugin installed. You will see entire file changes right after you lint. 

## Expected Behavior

The only thing that should happen when this lint rule is enabled `Footnotes -> Move Footnotes to the Bottom` is that all footnotes should always move to the bottom. Any inline block content should remain completely unaffected and unchanged.

## Screenshots

Image

#### Example:

Before running lint with value 'Footnotes -> Move Footnotes to the Bottom'

Image

AFTER running lint:

Image

## Device

- [x] Desktop
- [ ] Mobile

### Files Attached

data.json for all values used: [data.json](https://github.com/user-attachments/files/24192069/data.json)
Example file with issue: 

[OBSIDIAN SYNTAX.md](https://github.com/user-attachments/files/24192113/OBSIDIAN.SYNTAX.md)

vaaast-lake:
Fixes #1439

When the \`Move Footnotes to the Bottom\` rule ran, inline code values in the body and footnotes could be restored in the wrong locations. The issue was caused by \`ignoreListOfTypes()\` using shared placeholders for each ignore type and then restoring values sequentially after the document had been reordered.

This change promotes each ignored placeholder instance to a unique token and restores ignored values using a 1:1 token-to-original mapping. That makes restoration independent of document reordering.

Changes Made:
- Updated \`ignoreListOfTypes()\` to promote base placeholders to unique tokens after each ignore type is applied
- Changed restoration to use token-based replacement instead of sequential shared-placeholder replacement.
- Added regression tests covering inline code restoration in body content, inside footnotes, mixed body/footnote cases, and the `includeBlankLineBetweenFootnotes` option
- Added comments explaining the token promotion logic and related issue context

### DIFF
--- a/src/utils/ignore-types.ts
+++ b/src/utils/ignore-types.ts
@@ -1,74 +1,166 @@
-import {obsidianMultilineCommentRegex, tagWithLeadingWhitespaceRegex, wikiLinkRegex, yamlRegex, escapeDollarSigns, genericLinkRegex, urlRegex, anchorTagRegex, templaterCommandRegex, footnoteDefinitionIndicatorAtStartOfLine} from './regex';
-import {getAllCustomIgnoreSectionsInText, getAllTablesInText, getPositions, MDAstTypes} from './mdast';
-import type {Position} from 'unist';
-import {replaceTextBetweenStartAndEndWithNewValue} from './strings';
+import {
+  obsidianMultilineCommentRegex,
+  tagWithLeadingWhitespaceRegex,
+  wikiLinkRegex,
+  yamlRegex,
+  escapeDollarSigns,
+  genericLinkRegex,
+  urlRegex,
+  anchorTagRegex,
+  templaterCommandRegex,
+  footnoteDefinitionIndicatorAtStartOfLine,
+} from "./regex";
+import {
+  getAllCustomIgnoreSectionsInText,
+  getAllTablesInText,
+  getPositions,
+  MDAstTypes,
+} from "./mdast";
+import type { Position } from "unist";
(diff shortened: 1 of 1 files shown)


vaaast-lake:
Closing in favor of a new PR from the correct branch.

PR #1501, opened 2026-04-10
Fix: Restore Inline Code Correctly When Moving Footnotes to the Bottom

### LINKED ISSUE #1439: Bug: Footnotes -> Move Footnotes to the Bottom is incorrectly updating/modifying and moving INLINE CODE values

vaaast-lake:
Fixes #1439

When the `Move Footnotes to the Bottom` rule ran, inline code values in the body and footnotes could be restored in the wrong locations. The issue was caused by `ignoreListOfTypes()` using shared placeholders for inline code and then restoring values sequentially after the document had been reordered.

This change promotes ignored inline code placeholders to unique tokens and restores those values using a 1:1 token-to-original mapping. Other ignore types continue to use their existing placeholder-based flow so their intermediate formatting contracts remain unchanged.

Changes Made:
- Updated `ignoreListOfTypes()` to promote ignored inline code placeholders to unique tokens while preserving the existing placeholder flow for other ignore types
- Changed inline code restoration to use token-based replacement instead of sequential shared-placeholder replacement
- Added regression tests covering inline code restoration when moving footnotes to the bottom
- Added ignore list test coverage for unique inline code token promotion

### DIFF
--- a/src/utils/ignore-types.ts
+++ b/src/utils/ignore-types.ts
@@ -40,6 +40,7 @@ export const IgnoreTypes: Record<string, IgnoreType> = {
 export function ignoreListOfTypes(ignoreTypes: IgnoreType[], text: string, func: ((text: string) => string)): string {
   let setOfPlaceholders: {replacements: IgnoredReplacement[]}[] = [];
   const ignoreSessionPrefix = createIgnoreSessionPrefix();
+  let tokenIndex = 0;
 
   // replace ignore blocks with their placeholders
   let replaceValues: string[] = [];
@@ -53,11 +54,9 @@ export function ignoreListOfTypes(ignoreTypes: IgnoreType[], text: string, func:
       [replaceValues, text] = ignoreFunc(text, ignoreType.placeholder);
     }
 
-    // Promote base placeholders to unique tokens so ignored values can still be restored after rules reorder parts of the document, such as moving footnotes.
-    // see https://github.com/platers/obsidian-linter/issues/1439
-    const replacements = replaceValues.map((value: string, index: number) : IgnoredReplacement => {
+    const replacements = replaceValues.map((value: string) : IgnoredReplacement => {
       return {
-        token: createIgnoreToken(ignoreSessionPrefix, index),
+        token: createUniqueToken(ignoreType.placeholder, ignoreSessionPrefix, tokenIndex++),
         original: value,
       };
     });
@@ -90,8 +89,13 @@ function createIgnoreSessionPrefix(): string {
   return Math.random().toString(36).slice(2, 10);
 }
 
-function createIgnoreToken(ignoreSessionPrefix: string, index: number): string {
-  return `IGNORE_TOKEN_${ignoreSessionPrefix}_${index}`;
+function createUniqueToken(placeholder: string, ignoreSessionPrefix: string, index: number): string {
+  if (placeholder.startsWith('{') && placeholder.endsWith('}')) {
+    // Make brace-format placeholders unique per occurrence so that restoration
+    // is order-independent after rules reorder content. See: #1439
+    return placeholder.slice(0, -1) + `_${ignoreSessionPrefix}_${index}}`;
+  }
+  return placeholder;
 }
 
 /**
--- a/src/utils/ignore-types.ts
+++ b/src/utils/ignore-types.ts
@@ -53,6 +53,8 @@ export function ignoreListOfTypes(ignoreTypes: IgnoreType[], text: string, func:
       [replaceValues, text] = ignoreFunc(text, ignoreType.placeholder);
     }
 
+    // Promote base placeholders to unique tokens so ignored values can still be restored after rules reorder parts of the document, such as moving footnotes.
+    // see https://github.com/platers/obsidian-linter/issues/1439
     const replacements = replaceValues.map((value: string, index: number) : IgnoredReplacement => {
       return {
         token: createIgnoreToken(ignoreSessionPrefix, index),
--- a/src/utils/ignore-types.ts
+++ b/src/utils/ignore-types.ts
@@ -4,13 +4,16 @@ import type {Position} from 'unist';
 import {replaceTextBetweenStartAndEndWithNewValue} from './strings';
 
 export type IgnoreFunction = ((text: string, placeholder: string) => [string[], string]);
-export type IgnoreType = {replaceAction: MDAstTypes | RegExp | IgnoreFunction, placeholder: string};
+export type IgnoreType = {replaceAction: MDAstTypes | RegExp | IgnoreFunction, placeholder: string, useUniqueTokens?: boolean};
 export type IgnoredReplacement = {token: string, original: string};
+type IgnoredReplacementGroup =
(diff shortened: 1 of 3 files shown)


pjkaufman:
Hey @vaaast-lake , did you intend to close both of the PRs you opened?

vaaast-lake:
@pjkaufman Yes — #1500 was opened against the wrong branch, and I closed #1501 after confirming the failing test cases. I’m planning to work on a solution and reopen a PR with a proposed fix within next week. Would that be alright?

pjkaufman:
I believe that should be fine. You can run the test suite locally to see if things work as well.

src/utils/ignore-types.ts
```diff
@@ -4,12 +4,16 @@ import type {Position} from 'unist';
 import {replaceTextBetweenStartAndEndWithNewValue} from './strings';
 
 export type IgnoreFunction = ((text: string, placeholder: string) => [string[], string]);
-export type IgnoreType = {replaceAction: MDAstTypes | RegExp | IgnoreFunction, placeholder: string};
+export type IgnoreType = {replaceAction: MDAstTypes | RegExp | IgnoreFunction, placeholder: string, useUniqueTokens?: boolean};
+export type IgnoredReplacement = {token: string, original: string};
```
pjkaufman, line 8: What would be the point of not using unique tokens everywhere?

src/utils/ignore-types.ts
```diff
@@ -51,20 +56,48 @@ export function ignoreListOfTypes(ignoreTypes: IgnoreType[], text: string, func:
       [replaceValues, text] = ignoreFunc(text, ignoreType.placeholder);
     }
 
-    setOfPlaceholders.push({replacedValues: replaceValues, placeholder: ignoreType.placeholder});
+    // Make brace-format placeholders unique per occurrence so that restoration is order-independent after rules reorder content.
+    // see https://github.com/platers/obsidian-linter/issues/1439
+    if (ignoreType.useUniqueTokens) {
+      const replacements = replaceValues.map((value: string) : IgnoredReplacement => {
+        return {
+          token: `${ignoreType.placeholder}_${tokenIndex++}`,
```
pjkaufman, line 64: Why not change the the value to make sure it is inside the brace instead of after it?

But using numbers like this could cause clashes. So I am curious how this works when multiple calls are made to the ignore logic.

pjkaufman:
It may be best to try to use `crypto.randomUUID()` if it works on both desktop and mobile. That is probably the easiest way to create the unique identifier.

pjkaufman:
I am taking a look at another way of doing this, but it will likely take me a bit to get it ready.

pjkaufman:
Well, I think I made the change. Sorry it took so long and that it came after you made your own change: https://github.com/platers/obsidian-linter/pull/1512

It looks like it works. I used your added tests from here and decided to clean things up a decent bit from what you had done. I swapped a decent bit of how things were being handled, so I just created my own branch.

If you want to let me know if this works for your case, you can use the following zip: 
[main.zip](https://github.com/user-attachments/files/27230233/main.zip)

I will merge the other PR in if I don't hear back by Sunday.

Again, sorry about the confusion and putting you through the process of creating a PR only to have me create my own.

src/utils/ignore-types.ts
vaaast-lake: I did look at broader tokenization at first, but the full suite exposed a lot of regressions because some rules still relied on the current intermediate placeholder shape.

My intent in this PR was to keep the scope as small as possible and target the specific order-dependent case that was actually breaking here: Move Footnotes to the Bottom restoring ignored inline code after reordering the document. That was why I narrowed the change to inlineCode rather than applying unique tokens across all ignore types immediately.

Looking at #1512 now, it seems you took the broader unique-placeholder route and then updated the affected rules to understand the new form, which addresses the larger compatibility problem more directly.

src/utils/ignore-types.ts
vaaast-lake: I did try the brace-inside form earlier while iterating on this, but that ran into failures with space-between-chinese-japanese-or-korean-and-english-or-numbers, since that rule was still matching the exact placeholder literal up to the closing brace.

In the current branch here, I narrowed the change and only apply unique tokenization to inlineCode, while preserving the existing placeholder flow for other ignore types. The token form in this branch is the brace-outside variant, mainly to keep the original placeholder literal visible to code paths that still depend on it.

For the numbering concern: in this implementation I treated the generated values as invocation-local, since they are created and restored within the same ignoreListOfTypes() call, so they only need to be unique within that scope.

Looking at #1512, it seems to solve the same class of problem more directly by making the placeholder logic unique more generally and then updating the affected rules to understand that format, which is a cleaner general solution than the narrower compatibility-oriented approach I took here.

vaaast-lake:
@pjkaufman  

Thanks for following up on this and for taking the time to rework it in #1512.

Looking back on it, I think my understanding of some of the placeholder and rule interactions was still incomplete, especially once the change had to hold up across the full suite and not just the original reproducer. I tested the updated version on my side, and it looks good for the cases I was using.

I learned a lot from working through this issue and from seeing the approach you took here, so thank you for that as well.

And thank you again for continuing to maintain this project. I use it heavily, and I really appreciate the time and care that goes into keeping it working well.

pjkaufman:
Thanks for the swift reply and taking a look at this.

I am closing this in favor of #1512.

PR #1513, opened 2026-05-04
fix(emphasis-style): properly ignore math blocks

### LINKED ISSUE #1503: Bug: linter misinterpets underscores in math block as italics

## How to Reproduce

1. Create a blank note and add the following markfown
```md
adafsd$$
\begin{align}
\text{asfsss}_{}{d_{a}} 
\end{align}
$$
```

2. Run the linter

The linter replaces the underscores with asterisks when it shouldn't as this is inside dollar signs.

## Expected Behaviour

The linter should treat maths blocks as text in this circumstance.

## Device

- [x] Desktop
- [ ] Mobile

## Additional Context

This requires the emphasis setting to prefer astrisks.

Image

I suspect the line `adafsd$$` is being interpreted as the text `adafsd` followed by an empty inline math block `$$` which is not how this is parsed by Obsidian.

mitre88:
## Summary

The emphasis-style rule was incorrectly treating content inside math blocks as emphasis candidates.

## Root Cause

The rule had duplicate IgnoreTypes.math entries and the markdown parser incorrectly parses certain patterns like adafsd$$ as adafsd + empty inline math block.

## Fix

Ensures math and inlineMath ignore types are properly applied.

## Testing

Added test cases for math blocks with underscores.

Fixes #1503

### DIFF
--- a/__tests__/emphasis-style.test.ts
+++ b/__tests__/emphasis-style.test.ts
@@ -16,5 +16,45 @@ ruleTest({
       `,
       options: {style: 'asterisk'},
     },
+    {
+      // accounts for https://github.com/platers/obsidian-linter/issues/1503
+      // Math blocks with underscores inside should not be converted to emphasis
+      testName: 'Make sure math blocks are unaffected',
+      before: dedent`
+        adafsd$$
+        \begin{align}
+        \text{asfsss}_{}{d_{a}}
+        \end{align}
+        $$
+      `,
+      after: dedent`
+        adafsd$$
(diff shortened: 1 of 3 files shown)


pjkaufman:
@mitre88,  the UTs are not passing on this.

mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (failing-checks) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

PR #1522, opened 2026-05-11
Settings: Adopt Setting Definitions

liamcain:
- [x] Adopt setting definitions
- [x] Migrate all custom Suggest components to use the Obsidian API AbstractInputSuggest
- [x] Fix build requirements (added missing dependency, fixed prod trying to build integration tests)


pjkaufman:
The changes look pretty good so far. I just need to do the following yet before it can be merged:
- Add language text and keys for the new text
- Get UTs passing

pjkaufman:
I think this looks good. Assuming UTs pass, I will try to test it out tomorrow or later today and then merge it.

pjkaufman:
Well, just doing a quick glance, I hit a problem I need to iron out: 
image

I will look into fixing spacing and component position later.

liamcain:
where is that?

liamcain:
oh, found it. Custom regex replacement. I can migrate that too. I think these should all use a form approach instead of a growing list. It's easier to adapt to mobile, easier to validate, and generally simpler.

so: change + button to show a new FormModal (I added a class for that). then input the 4 values, press 'save' and then show a new row

Then the code can reuse the existing SettingGroup#onReorder and SettingGroup#onDeleteItem which will save a bunch of logic

pjkaufman:
Yeah. That is custom regex and command. I do think that the growing list males sense. I plan to move list options which are currently a newline or other separator based approach to the same kind of setting where the setting is a list rather than something inline. It's cleaner with the new page approach.

pjkaufman:
I think it should be pretty simple to do so. I am going to take a stab at it and see how it goes.

pjkaufman:
They also need an edit option which doesn't look too bad to add.

pjkaufman:
I think I am hitting a bit of wall here. So I am going to stop for now.

I was able to port the logic over for add for custom command. But now I am hitting two problems:
1. There is no option for editing an existing item directly
2. There is no option to change the display of an entry in a list to indicate whether it is disabled or enabled

Custom regex and Obsidian custom commands allow you to disable and enable them once added. This is more of a nice to have than a must have. But ideally it would stick around.

All current list based items allow you to edit them in the settings where they are. This is pretty much a must have as some of these settings take time to re-input so it is best to go ahead and keep that from being something that the user needs to do manually again because one letter in the custom regex or replacement isn't quite right.

pjkaufman:
Also, as a heads up, I am available tomorrow and possibly Friday to continue working on this, but I may not be able to do any coding between Friday and the 25th as I will be on vacation without my computer. So I may be able to do some minor editing after Friday, but I will mostly only be able to respond to messages.

pjkaufman:
I was able to add the logic for enabling and disabling the custom regex, but I still have to finish that for the custom commands. Once that is done, I will take a closer look at some of the settings. If all goes well at that point, I will go ahead and make a BRAT release for the initial changes.

image

pjkaufman:
I was able to get most of the custom command editing and enabling setup. However there is an issue where initializing the suggested is still showing a dropdown
I am trying to have it jot show that on initialization.

pjkaufman:
Ok. I think that is expected as the focus is on the suggester. So I believe the dropdown is expected. I will check with what it looks like in order to add a BRAT release now.

PR #1546, opened 2026-07-07
Fix Capitalize Headings incorrectly capitalizing words following a word starting with an ignored character.

### LINKED ISSUE #1544: Bug: capitalization errors in headings with parens

## Describe the Bug

Capitalization seems to be broken after an open paren '(' in a heading.

## How to Reproduce

1. In Linter settings, turn on "Lint on Save", then under the 'H1' tab, turn on "Headings should be formatted with capitalization," "Title Case".
2. Type a heading with one of the below and press cmd-S to trigger Linter

Observe:

``` markdown
# Long Way (Known good)
```

Capitalizes **wrongly** to

``` markdown
# Long Way (Known gOod)
```

but remove the parens:

``` markdown
# Long Way - Known good
```

capitalizes **correctly** to

``` markdown
# Long Way - Known Good
```

Note that this happens whether the closing paren is present or not.

## Expected Behavior

Adding parens should not impact capitalization of subsequent words.

``` markdown
# Long Way (Known good)
```

should capitalize to

``` markdown
# Long Way (Known Good)
```

## Device

- [x] Desktop (Mac)
- [ ] Mobile

I verified that happens in a test vault with no other plugins running.

asherhe:
Fixes #1544

When Capitalize headings encountered a word that starts with an ignored character (`'"(‘“-`), all subsequent words would invariably have their second character capitalized, regardless of whether they start with an ignored character as well.

The cause of this was the index to capitalize persisting across words. Ideally, it should be reset to 0 before checking for ignored characters again on the next word.

**Changes Made:**
- Reset the index to capitalize on a word-by-word basis
- Added test cases for the issue

### DIFF
--- a/__tests__/capitalize-headings.test.ts
+++ b/__tests__/capitalize-headings.test.ts
@@ -224,5 +224,17 @@ ruleTest({
         style: 'First letter',
       },
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1544
+      testName: 'Make sure that words in parentheses are correctly capitalized',
+      before: dedent`
+        # Long Way (Known good)
+      `,
+      after: dedent`
+        # Long Way (Known Good)
+      `,
+      options: {
+        style: 'Title Case',
+      },
+    },
   ],
 });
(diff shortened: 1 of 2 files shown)


pjkaufman:
Thank you for opening this. I will hopefully be able to verify the issue later today and then see about getting this merged.

pjkaufman:
I can reproduce locally, so this looks good. Thanks for tackling this.

PR #1548, opened 2026-07-12
fix: preserve dollar signs in YAML values when sorting keys

### LINKED ISSUE #1532: Bug: YAML Key Sort removed dollar ($) signs from YAML property values
- [X] I have verified that I am on the latest version of the Linter

## Describe the Bug

In a YAML property with a sequence of dollar signs ($), the YAML Key Sort linter will half the number of signs in a sequence every time it lints.
This continues multiple times, until there is a maximum of 1 dollar sign in a sequence.
The YAML Key Sort linter will half the dollar signs ($) from a YAML property, every time it lints.

## How to Reproduce

1. Create a note with a single YAML frontmatter propperty.
2. Add any number of `$` in it.
3. Lint the file, and observe the number of signs be halved.
4. This repeats every lint until the number of $ signs reaches 1.

This occurs for _every_ sequence of dollar signs within a single property.

``` markdown
---
my_property: $$$$$$$$$$
my_property2: $$$$ text in between $$$
---

```

This will become:

``` markdown
---
my_property: $$$$$
my_property2: $$ text in between $$
---
```

And after a second lint:

``` markdown
---
my_property: $$$
my_property2: $ text in between $
---
```
Third lint:
```markdown
---
my_property: $$
my_property2: $ text in between $
---
```
After the 4th lint this example will settle on:
```markdown
---
my_property: $
my_property2: $ text in between $
---
```

## Expected Behavior

The number of dollar signs in the property value should be unchanged, after any number of lints.

``` markdown
---
my_property: $$$$$$$$$$
my_property2: $$$$ text in between $$$
---

```

## Device

- [x] Desktop
- [ ] Mobile

## Additional Context

I use dollar signs in properties in my notes about restaurants, to indicate the general price range of the restaurant.

I'll typically use between one and three signs. But if I use more than one, then the linter will remove the signs each time it lints, giving me the wrong value.

Thank you for your time.
Kind regards.

ymxlx:
## What

Fixes #1532 — the YAML Key Sort rule silently halves runs of `$` in property values on every lint (`$$$$` → `$$` → `$`), losing data.

## Root cause

Not the `setYamlSection` regex suspected in the issue thread — that path uses a *function* replacement, which is `$`-safe. The actual culprit is in `YamlKeySort.getTextWithNewYamlFrontmatter`:

```ts
return text.replace(oldYaml, newYaml);
```

Here `newYaml` is used as a **replacement string**, so `String.prototype.replace` interprets `$$` as an escaped `$` — collapsing every `$$` to `$`. That's the halving.

## Fix

The codebase already solved this exact problem: `formatYAML()` in `src/utils/yaml.ts` wraps its replacement in `escapeDollarSigns()`. This applies the same existing helper in `yaml-key-sort.ts`:

```ts
return text.replace(oldYaml, escapeDollarSigns(newYaml));
```

One import + one call — no new utility, matching the established pattern.

## Test

Added a regression case to `__tests__/yaml-key-sort.test.ts`: a value `$$$$` must survive a key sort. Verified it **fails on the current code** (`$$$$` → `$$`) and **passes with the fix**.

## Verification

- New test fails before the change, passes after.
- Full suite green: **59 suites / 1196 tests**.
- eslint clean on both changed files.

## Note (not fixed here, to keep scope tight)

`src/rules/yaml-title-alias.ts:121` has the same `text.replace(oldYaml, newYamlValue)` pattern and would corrupt `$`-runs in aliases the same way. I left it out of this PR to keep the fix focused on #1532 — happy to send a follow-up (or fold it in here if you'd prefer).

### DIFF
--- a/__tests__/yaml-key-sort.test.ts
+++ b/__tests__/yaml-key-sort.test.ts
@@ -359,5 +359,23 @@ ruleTest({
         yamlSortOrderForOtherKeys: 'Ascending Alphabetical',
       },
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1532
+      testName: 'Dollar signs in YAML values are preserved when sorting keys',
+      before: dedent`
+        ---
+        zeta: last
+        alpha: $$$$
+        ---
+      `,
+      after: dedent`
+        ---
+        alpha: $$$$
+        zeta: last
+        ---
+      `,
+      options: {
+        yamlSortOrderForOtherKeys: 'Ascending Alphabetical',
+      },
+    },
   ],
 });
--- a/src/rules/yaml-key-sort.ts
+++ b/src/rules/yaml-key-sort.ts
@@ -2,6 +2,7 @@ import {Options, RuleType} from '../rules';
 import RuleBuilder, {BooleanOptionBuilder, DropdownOptionBuilder, ExampleBuilder, OptionBuilderBase, TextAreaOptionBuilder} from './rule-builder';
 import dedent from 'ts-dedent';
 import {parseYAML, getYAMLText, loadYAML, setYamlSection, astToString, getEmptyDocument} from '../utils/yaml';
+import {escapeDollarSigns} from '../utils/regex';
 import {Document} from 'yaml';
 import {YamlCSTTokens, YamlNode} from '../typings/yaml';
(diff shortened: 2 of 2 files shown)


pjkaufman:
This looks good. Thanks for taking the time to take a look and fix this.

PR #1560, opened 2026-08-03
fix(trailing-spaces): stop inserting a literal "$1" on whitespace-only lines

### LINKED ISSUE #1542: Bug: trailing-spaces with two-space-line-break inserts literal '$1' for single-space-only lines
## Describe the bug

When `trailing-spaces` is enabled with `two-space-line-break: true`, a line consisting of **exactly one space** gets replaced with the literal string `$1` instead of being cleaned up.

## Steps to reproduce

1. Enable `trailing-spaces` with `two-space-line-break: true`
2. Create a note containing a line with exactly one space character:

```markdown
# Heading

%% some comment %%
 
# Next heading
```

(The blank line between `%%` and `# Next heading` is a single space, not truly empty.)

3. Run Linter on the file

## Expected behavior

The single-space line should be cleaned up (reduced to an empty line).

## Actual behavior

A literal `$1` appears in place of the single-space line:

```markdown
# Heading

%% some comment %%
$1
# Next heading
```

## Root cause

In `src/rules/trailing-spaces.ts`, Block 3 (empty/whitespace-only lines), patterns handling `twoSpaceLineBreak`:

```ts
// Pattern #10 — BUG: no capture group, $1 is literal
/^[ \t]$/gm   →   "$1"

// Pattern #11 — same bug
/^[ \t]{3,}$/gm   →   "$1"
```

Both regexes use `[ \t]` (character class) instead of `(...)` (capture group), so `$1` in the replacement references a non-existent group. In JavaScript, `String.replace()` treats `$1` as a literal string when the regex has no capture groups.

Compare with the correctly-written patterns #2–#4 which use `(\S)` as a capture group:

```ts
/(\S)[ \t]$/gm   →   "$1"   // correct: capture group exists
```

## Suggested fix

Change replacements for patterns #10 and #11 from `"$1"` to `""` (empty string), since these handle whitespace-only lines that should be reduced to nothing.

## Environment

- Linter version: 1.32.0
- Obsidian version: latest

MsfPablo:
## Summary

Fixes #1542. With `twoSpaceLineBreak` enabled, `TrailingSpaces` used `text.replace(/^[ \t]$/gm, '$1')` and `text.replace(/^[ \t]{3,}$/gm, '$1')` to clear whitespace-only lines of 1 or 3+ characters. Neither regex has a capture group, so `'$1'` is treated as a literal replacement string rather than a backreference — a whitespace-only line gets replaced with the literal text `$1` instead of being emptied.

Changed the replacement string from `'$1'` to `''` in those two calls only. Left the third call on the next line (`text.replace(/^( ?\t\t? ?)$/gm, '$1')`, which handles exactly-two-whitespace lines) untouched — it has a real capture group and is working as intended, out of scope for this issue.

## Test plan

- Reproduced the bug before fixing: `"text\n \nmore text\n    \nend"` → `"text\n$1\nmore text\n$1\nend"` on current `master`
- Added two regression tests to `__tests__/trailing-spaces.test.ts` covering 1-space and 3+ space whitespace-only lines with `twoSpaceLineBreak` enabled, plus confirmed the existing "two spaces preserved" case still passes
- Confirmed the new tests fail without the fix (stashed the source change and reran) and pass with it
- `npx jest __tests__/trailing-spaces.test.ts` — 15/15 pass
- Full suite (`npm test`), `npm run build`, `npm run docs`, and `eslint . --ext .ts` all clean; `npm run docs` regenerates with zero diff

### DIFF
--- a/__tests__/trailing-spaces.test.ts
+++ b/__tests__/trailing-spaces.test.ts
@@ -192,5 +192,45 @@ ruleTest({
         ${''}
       `,
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1542
+      testName: 'Whitespace only lines are emptied instead of being replaced with a literal "$1" when two space line break is enabled',
+      before: dedent`
+        Some text
+        ${' '}
+        More text
+        ${'   '}
+        Even more text
+        ${'\t'}
+        The end
+      `,
+      after: dedent`
+        Some text
+        ${''}
+        More text
+        ${''}
+        Even more text
+        ${''}
+        The end
+      `,
+      options: {
+        twoSpaceLineBreak: true,
+      },
+    },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1542
+      testName: 'A whitespace only line of exactly two spaces is still preserved when two space line break is enabled',
+      before: dedent`
+        Some text
+        ${'  '}
+        More text
+      `,
+      after: dedent`
+        Some text
+        ${'  '}
+        More text
+      `,
+      options: {
+        twoSpaceLineBreak: true,
+      },
+    },
   ],
 });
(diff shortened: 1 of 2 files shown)


pjkaufman:
This looks good. I will go ahead and get this merged. Thanks for fixing this @MsfPablo .

Issue #1502, opened 2026-04-21
FR: Be able to ignore tags that start with a number


thelegend09:
## Is Your Feature Request Related to a Problem? Please Describe.

Sometime I make lists and I number things using a hashtag. `The #1 thing to do is...`. Obsidian does NOT mark that as a tag, and I do NOT mean it as a tag. However, the linter sees it as a tag and moves it in the tags frontmatter when the setting `Move all tags to YAML frontmatter of the document.` is selected.

## Describe the Solution You'd Like

I would like Obsidian Linter to ignore that tags that start with a number. It would helpful in such situations: 

``` markdown
the #1 thing to do is... (enumeration)
```

``` markdown
Please call 999 999-9999 #24534 (phone extention)
```

``` markdown
I live on 123, Acme Street apt. #3 (apartment number)
```

## Describe Alternatives You've Considered

I added a bunch of number in the `Tags to ignore` box. It will do for now, but it is clearly a workaround. 

Thanks for this wonderful plugin, I really depend on it!!!

isukennedy:
I definitely concur with this suggestion.  It creates a conflict with metadata menu and it's automatic table generation.

pjkaufman:
The change for this has been merged as a part of aligning with the Obsidian spec. So I am going to close this. I am not sure it ignores all tags that start with a number unless that is what the Obsidian spec is (Note: I did not work on the fix, so I am just including this due to my ignorance on this).

Issue #456, opened 2022-10-23
Bug: Two lines around Code Fences.


jerviscui:
## Describe the Bug

Will add two lines around Code Fences.

## How to Reproduce

1. Enable **Empty Line Around Code Fences**
2. Enable **Consecutive blank lines**

## Expected Behavior

Expected output if applicable:

The formatted result:

~~~
``` markdown
1. 修改 php.ini


    ```
    upload_max_filesize = 20M
    post_max_size = 20M
    ```


2. 修改 Nginx 配置文件


    ```
    client_max_body_size    20m; # 文件最大大小
    ```


1111

```
int a = 0;
```

2222
```
~~~

The code block not in the list is correct, it is around with a blank line.
The code block in the list will be around two blank lines. I think the code block around with one line is correct.

## Device

- [x] Desktop

jerviscui:
Oh. I just to show the content of the file, so use the markdown block.

~~~
``` markdown 
~~~

pjkaufman:
I will see about reproducing this tonight and if I can, I will aim to have it done for the next release.

pjkaufman:
Looks like I have a fix for the bug. The issue is that it was trying to match the whitespace of it with that of the previous line which is technically how it was written, but I see that that is not really necessary, so I have gone ahead and removed that part of the logic. This should take care of the problem and will hopefully not create any others.

pjkaufman:
This should be taken care of now. It will be pushed out with the next release. If you would like, you may go ahead and use the master version of the plugin before then.

Please let us know if the issue persists either on master or in the next release.

PR #1561, opened 2026-08-03
fix(regex): purely-numeric #tags no longer match, per Obsidian's tag spec

### LINKED ISSUE #1535: move-tags-to-yaml harvests purely-numeric "#tags" (#149, #1) that Obsidian itself rejects
**Describe the bug**

"Move tags to YAML" moves purely-numeric hashtags from the note body into the frontmatter `tags:` list — e.g. `#149`, `#193`, `#1`. Obsidian itself does *not* treat these as tags (a valid tag must contain at least one non-numerical character — `#1984` is invalid, `#y1984` is valid; see the [tag format docs](https://help.obsidian.md/Editing+and+formatting/Tags#Tag+format)). So GitLab MR refs (`#149`) or issue refs (`#1`) written in prose get harvested into `tags:` as YAML **integers** — they render as red numbers in the Properties UI and never appear in the tag index.

**Root cause**

`tagWithLeadingWhitespaceRegex` in `src/utils/regex.ts`:

```
/(\s|^)(#[\p{L}\-_\d/\p{Emoji_Presentation}]+)/gu
```

The character class includes `\d` with no requirement of a non-numeric character, so it matches `#149`. This diverges from Obsidian's own rule that a tag must contain at least one non-numerical character.

**To reproduce**

1. Enable "Move tags to YAML".
2. In a note body, write `Deploy — MR #149, merged`.
3. Lint the file.
4. `149` is added to frontmatter `tags:` (as an integer); with "Remove whole tag" enabled, `#149` is also stripped from the body.

**Expected behavior**

Purely-numeric hashtags should not be treated as tags (matching Obsidian's tag spec): `#149` is left untouched in the body and never added to frontmatter.

**Proposed fix**

Require at least one non-numeric character via a lookahead:

```
/(\s|^)(#(?=[\p{L}\-_\d/\p{Emoji_Presentation}]*[\p{L}\-_/\p{Emoji_Presentation}])[\p{L}\-_\d/\p{Emoji_Presentation}]+)/gu
```

This keeps valid tags like `#26-1-pathfinder`, `#y1984`, `#2024-2025`, `#ai/governance` while excluding `#149`, `#1`, `#193`.

**Environment**

- Linter version: 1.31.x
- Obsidian tag format: https://help.obsidian.md/Editing+and+formatting/Tags#Tag+format

MsfPablo:
## Summary

Fixes #1535. `tagWithLeadingWhitespaceRegex` in `src/utils/regex.ts` matched any `#` followed by one or more characters from `[\p{L}\-_\d/\p{Emoji_Presentation}]` — including tags made up entirely of digits, like `#123`. Per [Obsidian's tag spec](https://help.obsidian.md/Editing+and+formatting/Tags#Tag+format): "Tags must contain at least one non-numerical character. For example, `#1984` isn't a valid tag, but `#y1984` is." So `move-tags-to-yaml` (and, since this regex is shared, `matchTagRegex` and the `tag` `IgnoreTypes` filter in `ignore-types.ts`) were treating purely-numeric hashes as real tags when Obsidian itself doesn't.

Added a lookahead requiring at least one non-digit character in the tag body: `#(?=\d*[\p{L}\-_/\p{Emoji_Presentation}])[\p{L}\-_\d/\p{Emoji_Presentation}]+`. This only changes purely-numeric matches — anything with a letter, emoji, `-`, `_`, or `/` anywhere in it (e.g. `#y1984`, `#1984book`, `#inbox/to-read`) still matches exactly as before.

## Test plan

Validated the new regex against an 18-case table before committing: `#123`, `#1984`, `#0` correctly rejected; `#y1984`, `#1984book`, `#a1`, `#1_2`, `#camelCase`, `#kebab-case`, `#snake_case`, `#inbox/to-read`, `#🎉` etc. still match; `"#123 #456 #ok"` correctly yields only `#ok`.

- Added a regression test to `__tests__/move-tags-to-yaml.test.ts`; confirmed it fails without the fix and passes with it
- `npx jest __tests__/move-tags-to-yaml.test.ts __tests__/format-tags-in-yaml.test.ts` — 21/21 pass (format-tags-in-yaml needed no changes — it operates on YAML tags, not body `#tags`)
- Full suite (`npm test`), `npm run build`, `npm run docs`, and `eslint . --ext .ts` all clean; `npm run docs` regenerates with zero diff

One caveat I can't fully verify: `#12/34` and `#1-2` still match under this fix, since `/` and `-` are non-numerical characters per the literal wording of the spec — I believe that's correct, but couldn't confirm against Obsidian's actual parser for those specific inputs.

### DIFF
--- a/__tests__/move-tags-to-yaml.test.ts
+++ b/__tests__/move-tags-to-yaml.test.ts
@@ -384,5 +384,23 @@ ruleTest({
         howToHandleExistingTags: 'Remove whole tag',
       },
     },
+    { // accounts for https://github.com/platers/obsidian-linter/issues/1535
+      testName: 'Purely numeric hashes are not treated as tags since Obsidian requires at least one non-numerical character',
+      before: dedent`
+        Numeric only: #123
+        Year like: #1984
+        Valid tag with digits: #y1984
+        Valid tag ending in digits: #1984book
+      `,
+      after: dedent`
+        ---
+        tags: [y1984, 1984book]
+        ---
+        Numeric only: #123
+        Year like: #1984
+        Valid tag with digits: #y1984
+        Valid tag ending in digits: #1984book
+      `,
+    },
   ],
 });
--- a/src/utils/regex.ts
+++ b/src/utils/regex.ts
@@ -14,7 +14,8 @@ export const wikiLinkRegex = /(!?)\[{2}([^\][\n|]+)(\|([^\][\n|]+))?(\|([^\][\n|
 // based on https://davidwells.io/snippets/regex-match-markdown-links
 export const genericLinkRegex = /(!?)\[([^[]*)\](\(.*\))/g;
 // based on https://help.obsidian.md/Editing+and+formatting/Tags#Tag+format
-export const tagWithLeadingWhitespaceRegex = /(\s|^)(#[\p{L}\-_\d/\p{Emoji_Presentation}]+)/gu;
+// the lookahead enforces "tags must contain at least one non-numerical character", so `#1984` is not a tag while `#y1984` is
+export const tagWithLeadingWhitespaceRegex = /(\s|^)(#(?=\d*[\p{L}\-_/\p{Emoji_Presentation}])[\p{L}\-_\d/\p{Emoji_Presentation}]+)/gu;
 export const obsidianMultilineCommentRegex = /^%%\n[^%]*\n%%/gm;
 export const wordSplitterRegex = /[,\s]+/;
 export const whitespaceSplitterRegex = /\S+/g;


thelegend09:
This PR would also fix issue #1502

MsfPablo:
Good catch @thelegend09 — yes, #1561 resolves all three concrete examples in #1502 (`#1`, `#24534`, `#3`), since they're purely-numeric tags and the new lookahead rejects any tag with no non-numerical character: `#(?=\d*[\p{L}\-_/\p{Emoji_Presentation}])`.

One scoping note for review: #1502's title says "tags that *start with* a number", but all three of its examples are purely numeric. This PR aligns with Obsidian's own tag spec ("Tags must contain at least one non-numerical character"), so a tag like `#1984book` — starts with digits but contains a letter — still matches, since Obsidian itself treats that as a valid tag. If the broader "starts-with-a-digit" behavior is wanted separately that'd be a follow-up, but it would diverge from Obsidian's spec.

pjkaufman:
This looks fine to me. I don't necessarily like that the fix is in the regex rather than the match tag logic, but this should work as is. As such, I will proceed with this change. Thanks for creating a fix @MsfPablo .

PR #1566, opened 2026-08-21
fix: ordered list indicators become NaN when terminated with ')'

demrom:
### Bug

`1) a` becomes `NaN. a`.

`updateOrderedListItemIndicators` reads the item number from group 4 of

```
/^(( |\t|> )*)((\d+(\.|\)))|[-*+])([^\n]*)$/gm
```

which is `\d+(\.|\))` — the digits *with* their terminator — and coerces it. `Number('1.')` is `1` (JS accepts a trailing decimal point); `Number('1)')` is `NaN`, which then propagates down the rest of the list through `NaN + 1` under `ascending`.

Reachable via two documented settings: `preserve-start` enabled, or `number-style: preserve`. The loss is silent and unrecoverable — the source numbers no longer exist in the file. In one vault this destroyed 197 indicators across 7 notes before it was traced.

### Fix

`parseInt(_$4, 10)`; parsing stops at the terminator. Regex, group numbering and all other paths unchanged.

### Tests

One case per trigger in `__tests__/ordered-list-style.test.ts`. Both fail on master with `NaN. Item 1`, pass here. Full suite green (1199 tests / 59 suites); eslint clean.

### DIFF
--- a/__tests__/ordered-list-style.test.ts
+++ b/__tests__/ordered-list-style.test.ts
@@ -1,6 +1,7 @@
 import OrderedListStyle from '../src/rules/ordered-list-style';
 import dedent from 'ts-dedent';
 import {ruleTest} from './common';
+import {OrderListItemStyles} from '../src/utils/mdast';
 
 ruleTest({
   RuleBuilderClass: OrderedListStyle,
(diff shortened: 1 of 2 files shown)


pjkaufman:
This looks to be an existing issue. I don't see any issue around it, but I did verify that this does happen locally. I will go ahead and merge this. Thanks for creating a fix @demrom .

PR #1574, opened 2026-09-01
Add maximum file line count

### LINKED ISSUE #1573: FR: add max line number for linted files

i dont know if the issues exist, but linting on high line numbers counts (i have 1800 lines markdown files) it is quite laggy (~15s on my computer)

add max line number settings/preferences.

Ni-zav:
## Summary

- Add a setting to skip files above a line-count limit.
- Use `0` for no limit.
- Skip custom lint commands for skipped files.

Fixes #1573

## Tests

- `npm.cmd test -- --runInBand __tests__/rules-runner.test.ts --forceExit`
- ESLint on changed files
- `npm.cmd run build`


pjkaufman:
I would be fine with something like this being added so users can make any alterations to which file lengths to lint. There is a change coming that should speed things up dramatically from what I am seeing. So that might mitigate the need for this.

Ni-zav:
> I would be fine with something like this being added so users can make any alterations to which file lengths to lint. There is a change coming that should speed things up dramatically from what I am seeing. So that might mitigate the need for this.

thank you, waiting for that changes

PR #1589, opened 2026-09-14
Add CJK Spacing Missing Boundaries With Multiple Inline Math or Code on One Line Tests

### LINKED ISSUE #1554: Bug: Regression: CJK spacing misses boundaries when multiple inline math expressions are on the same line
> [!IMPORTANT]
> **AI assistance disclosure:** The text of this issue was written by AI based on the reproduction details and verification I provided. AI was also used to analyze the source code and Git history and to identify the suspected regression and proposed fix. The bug itself was discovered by me during actual use after updating Linter. I personally reproduced the behavior and verified the local fix described below.

## Describe the Bug

After updating Linter to version `1.32.0`, the rule `space-between-chinese-japanese-or-korean-and-english-or-numbers` no longer reliably inserts spaces around inline math when multiple inline math expressions occur on the same line.

A line containing one inline math expression is formatted correctly. When multiple expressions occur on the same line, some boundaries are skipped.

## How to Reproduce

1. Use Linter `1.32.0`.
2. Enable `Space between Chinese, Japanese or Korean and English or numbers`.
3. Create a note containing:

```markdown
测试$math$测试$math$测试
```

4. Run Linter.

## Actual Behavior

```markdown
测试 $math$测试$math$ 测试
```

Only the boundary before the first expression and the boundary after the second expression are handled.

## Expected Behavior

```markdown
测试 $math$ 测试 $math$ 测试
```

Every inline math expression adjacent to CJK text should have spaces added on both sides.

## Suspected Regression

The behavior appears to have been introduced by commit [`29ea801b`](https://github.com/platers/obsidian-linter/commit/29ea801b95dba0be59eaef6fae8e8f0ee9f17734), included in version `1.32.0`.

That commit changed the placeholder pattern from:

```ts
.replaceAll('}', '\\}')
```

to:

```ts
.replaceAll('}', '.+\\}')
```

Since `.+` is greedy, it can span multiple unique inline-math placeholders on the same line. Consequently, multiple adjacent placeholder boundaries can be treated as one match.

The previous release, `1.31.2`, used fixed placeholders and does not appear to have this specific behavior.

## Locally Verified Fix

I changed the bundled equivalent of:

```ts
.replaceAll('}', '.+\\}')
```

to:

```ts
.replaceAll('}', '[^}]+\\}')
```

After reloading the plugin, the reproduction case was formatted correctly. Restricting the generated placeholder ID to characters before the next `}` prevents a match from spanning multiple placeholders.

## Suggested Regression Test

The existing test suite contains a case for one inline math expression, but it would be useful to add a case containing multiple inline math expressions on the same line:

```ts
{
  testName: 'Make sure spaces are added around multiple inline math expressions on the same line',
  before: dedent`
    测试$math$测试$math$测试
  `,
  after: dedent`
    测试 $math$ 测试 $math$ 测试
  `,
},
```

This could be added to:

`__tests__/space-between-chinese-japanese-or-korean-and-english-or-numbers.test.ts`

## Environment

- Linter version: `1.32.0`
- Operating system: Windows
- Obsidian version: `1.12.7`

anandghegde:
Fixes #1554

Credit to @0kitasan for tracking this down to the placeholder regex and suggesting the `[^}]+` change in the issue. This PR applies it and adds regression tests.

Changes Made:
- Changed the placeholder pattern in `space-between-chinese-japanese-or-korean-and-english-or-numbers` from `.+\}` to `[^}]+\}`
- Added tests for multiple inline math expressions and multiple inline code blocks on the same line (the examples from the issue and its follow-up comment)

## Why this happens

Since #1512 every placeholder has a unique id, e.g. `{INLINE_MATH_PLACEHOLDERmx1abc23}`. To match "any id", the rule turned each `}` in the placeholder list into `.+\}`. `.+` is greedy, so on a line like

```
测试{INLINE_MATH_PLACEHOLDERid1}测试{INLINE_MATH_PLACEHOLDERid2}测试
```

a match that starts at the first placeholder runs on to the last `}` on the line. The head regex's one match eats both placeholders plus the `测试` between them, and so does the tail regex's. The result is that only the boundary before the first placeholder and after the last one get a space. `.` doesn't match newlines, which is why it only happens when the placeholders are on the same line. The same thing happens across placeholder types, since link, wiki link, inline math and inline code are all in one alternation.

## Why `[^}]+`

The id is generated between the placeholder name and its closing `}`, and it never contains a `}`. So `[^}]+` always stops at that placeholder's own closing brace and can't reach into the next one. I picked it over something like `[a-z0-9]+` because it doesn't depend on the id format in `getNewPlaceHolder`, and it still works if another rule changes the case (restoring placeholders already uses the `i` flag for that reason).

I didn't touch `blockquote-style.ts`. It uses the same `.+}` pattern, but only to check whether a line contains a placeholder (`match` used as a boolean), so greediness doesn't change its result there.

## Testing

- Added the two tests first. Both failed on current `master` with the output reported in the issue (`测试 $math$测试$math$ 测试`, and ``直接修改数据库 `library`表对应行的`exclusionPatterns` 字段``). Both pass with the change.
- `npx jest`: 59 suites, 1206 tests passed
- `npx eslint` on the two changed files: clean
- `npm run build`: succeeds


anandghegde:
Merged master to resolve the conflict with #1590. That PR removed the placeholder regex this PR changed, and the two regression tests for #1554 pass on its new approach, so this PR now only adds those tests.

pjkaufman:
@anandghegde , is this still an issue with what is on master?

anandghegde:
No, it's fixed on master. #1590's rewrite of the CJK spacing rule handles multiple inline math placeholders, so after merging master this PR only adds the two regression tests for #1554 (both pass). Happy to keep it as a tests-only PR, or close it if you'd rather not have them.

pjkaufman:
I'll double check the current tests and see if these cases are already covered. If so, we can close it. Otherwise, I am fine with adding them.

pjkaufman:
This should be fine to add.

PR #1590, opened 2026-09-14
feat: replace document masking with a protected-range index

saberzero1:
Linting a large note is currently dominated by work that has nothing to do with the rules
themselves. This branch removes that work. Output is byte-identical throughout.

## What changed

Rules used to be handed a **masked copy** of the document, with code blocks, YAML, links and so on
swapped for placeholder tokens, then have their result unmasked. Every rule that ignored a
construct rebuilt the document twice and parsed it again.

Rules now read the **original document** and consult a **protected-range index** that says which
regions they must not touch. Rules whose edits do not overlap share one snapshot, one parse and one
set of ranges, and their changes are applied together.

Nothing about which rules do what, or in what order, has changed.

## Results

Measured on an 866KB note with every rule enabled:

| | before | after |
|---|---:|---:|
| full-document parses | 119 | **15** |
| time in the parser | 181.7s | **8.2s** |
| total lint | 337.6s | **13.7s** |

**About 25x.** Both figures come from the same session on the same machine, which matters more than
it sounds: measured separately later, master's median over three runs was 381.0s and this branch's
median over five was 10.9s, which would read as 35x. Neither pairing is dishonest, but only the
first was measured under identical conditions, so that is the one quoted. The branch has genuinely
improved since — the cache work below accounts for most of it — but a cross-session ratio is not
evidence and is not claimed here.

Because that document is an outlier, the same comparison over generated notes of realistic size,
four documents per size, every rule enabled:

| size | before | after | ratio |
|---:|---:|---:|---:|
| 2KB | 125.1ms | 13.7ms | **9.11x** |
| 5KB | 341.1ms | 44.2ms | 7.72x |
| 10KB | 713.5ms | 87.1ms | 8.20x |
| 25KB | 1946.8ms | 253.3ms | 7.69x |
| 50KB | 4047.4ms | 583.0ms | 6.94x |
| 100KB | 8837.8ms | 1355.4ms | 6.52x |

**No size regressed and there is no crossover.** "Messy" documents that actually give the rules work
to do, rather than exercising the fast path, come out 6.97x at 10KB and 6.26x at 50KB.

Both sides of that table were measured in a single session under identical conditions, at the commit
immediately before this branch was rebased onto master; the rebase added only an upstream YAML fix
and a constructor arity correction. Re-measuring one point later gave 88.5ms where the table says
87.1ms, and the large fixture moved the other way, so treat run-to-run variance on this machine as
roughly ten percent and read the ratios rather than the absolute milliseconds.

The gain is largest on the *smallest* notes, which says the old cost was dominated by rebuilding the
document once per rule rather than by parsing.

## The editor path was the real bottleneck for an open note

Everything above measures `RulesRunner.lintText`. In Obsidian, "lint this file" on an **open editor**
then hands the result to CodeMirror through `updateEditor`, and that was costing far more than the
linting.

It walked the diff and dispatched **one CodeMirror transaction per chunk**, computing each position
by splitting the accumulated document on newlines. On the 866KB fixture:

| | |
|---|---:|
| `diff_main` | 83ms |
| position bookkeeping (4,548 whole-document splits) | **2,906ms** |
| the same walk with a running offset | **0.8ms** |
| transactions dispatched | **2,833 → 1** |

The diff already describes every change relative to the original text, and CodeMirror accepts an
array of changes in one transaction on exactly that basis. So the positions are computed with a
running offset and dispatched once.

This is why linting a large note could take minutes while the rules themselves took seconds — and
why it was invisible to us: no test could reach `updateEditor`. The mapping is now a pure function
with 24 tests plus 1,000 randomly generated document pairs, each diffed, mapped, applied and
compared, carriage returns and emoji included.

**`runLinterFile` never had this problem** — lint-on-file-change, folder lint and vault lint read,
lint and write without diffing, so they always received the full benefit of the work above.

One deliberate behaviour change: a selection spanning edited text maps slightly differently, because
one transaction maps a range once instead of threading it through thousands. Cursor positions are
unchanged and the document is identical. A single atomic undo step also seems the better model for
"lint this document".

## Memory

Both caches were bounded by entry count. The parse cache held 200 entries, and one entry for an
866KB document retains about 12MB, so a vault of large notes could have retained roughly **2.35GB**.
Linting a vault starts every file against these shared module-global caches and nothing ever clears
them.

They are now bounded by estimated retained bytes — 64MiB for parses, 16MiB for contexts — with a
per-entry ceiling so one pathological document cannot empty the cache. An oversized parse is still
returned to its caller, just not stored, so it degrades to parsing every time rather than failing.

| | before | after |
|---|---:|---:|
| retained, one 866KB lint | 219.5MB | **59.3MB** |
| worst case | ~2.35GB | **80MiB, bounded** |

Over 1,000 distinct notes with the caches never cleared, heap plateaus by about note 400 at 62.1MiB
and stays there.

`quick-lru` is replaced by `lru-cache`, which supports byte budgets natively. Bundle cost:
936,024 to 955,602 bytes, **+19,578 (2.09%)**.

## How the output is verified

`__tests__/rules-runner-differential.test.ts` lints a corpus built from the rules' own `examples[]`
plus adversarial cases, and compares a hash per document against a committed **225-entry snapshot**.
Any behavioural change to any rule fails it with the offending document named, and `DUMP_PATH=…`
writes the full text when you want to read the difference rather than just detect it.

**Every commit on this branch is byte-identical** against the output of the code as it stood before
any of this work.

Worth knowing: the *old* masking output is **not reproducible** — it injects random UUID
placeholders, so two runs of the same document could differ. Removing masking made it deterministic,
which is what makes a snapshot possible at all.

Both performance harnesses are optional and skip unless pointed at a large note:

```
LINTER_PERF_FIXTURE=/path/to/big-note.md LINTER_PERF_SETTINGS=/path/to/data.json \
  LINT_FULL_FIXTURE=1 npx jest __tests__/parse-count.test.ts
```

`__tests__/parse-count.test.ts` asserts a parse-count ceiling, so reintroducing redundant parses fails CI
rather than going unnoticed — which is exactly how the original regression survived.

## Four things for the maintainer

**1. This adds `patch-package` and a `postinstall` hook, and that is a decision for you.**
`patches/mdast-util-from-markdown+2.0.3.patch` backports upstream commit `53875a7`, which is merged
but still unreleased — `mdast-util-from-markdown` is at 2.0.3 on npm as of writing, and I re-checked.
The patch is **not** a performance win; it is carried so this project does not silently diverge from
a fix that is already upstream. Please delete both the patch and the `postinstall` step once a
release contains it. If you would rather not take a postinstall hook at all, the patch can be dropped
with no measurable effect.

**2. There is a latent mis-arity in `TextAreaOptionBuilder.buildOption`, and it has visible
consequences.** `Option`'s constructor takes five parameters; the call passes seven. So
`defaultValue` receives `null`, `ruleAlias` receives the joined default string, and the last two
arguments are discarded.

This branch **preserves that behaviour exactly** and only corrects the arity, verified by
regenerating the documentation on both sides: master and this branch both emit `` `null` `` for the
affected options. Changing what the settings pane shows is not a performance change and should not
ride along in a performance PR.

It does appear to be a real bug, and it is why `npm run docs` currently rewrites
`docs/docs/settings/{content,heading,yaml}-rules.md` to show `` `null` `` in place of real defaults
such as the capitalisation word lists. **That drift already exists on master** — regenerating there
rewrites `README.md` and the same three files, with `yaml-rules.md` moving by 520 lines against this
branch's 19. This branch therefore leaves the generated documentation untouched rather than
committing `null` over the real defaults, which would look like a documentation regression caused by
this PR. Fixing the arity properly would restore the true values, and is worth doing separately.

**3. Three pre-existing corruption bugs were fixed** where reproducing them would have required
reproducing the corruption: a negative offset when a heading contained `&amp;`, doubled whitespace
in list items, and dangling references when footnotes were renumbered. Each is verified to change
nothing across the corpus.

**4. The editor fix changes selection mapping slightly.** A selection spanning edited text maps once
through a single transaction instead of being threaded through thousands. Cursors land in the same
place and the document is identical, but if you have a strong opinion about selection behaviour
during a lint, that is the line to look at.

## Checks

`npm run compile` — build, docs, lint and test — passes end to end.

- **80 suites, 1,652 tests.** A clean checkout with no environment variables set — what CI does —
  runs 79 suites and 1,651 tests, skipping only the optional large-note performance harness.
- `npm run lint` (`eslint --fix`) makes no changes.
- `tsc --noEmit` reports 60 errors, all pre-existing; 28 of them are already in `src/main.ts`. The
  count was 63 before — removing `quick-lru` removed three errors originating in its own type
  declarations.
- Build passes.
- The only thing `compile` leaves modified is the generated rule documentation, for the pre-existing
  reason described above. It does the same on master.


CONTRIBUTING.md
```diff
@@ -746,25 +746,49 @@ At times you may find a function or variable that is needed in multiple rules. T
...
+- the **edit range**, the characters the rule rewrites;
+- the **guard range**, the characters that had to be visible for the rule to make that change at all. If the guard
+  range is protected, the change is skipped.
+
+For example, [Remove Space Before or After Characters](src/rules/remove-space-before-or-after-characters.ts)
+deletes whitespace next to a punctuation character the user configured, so **both** the whitespace and that character
+have to be unprotected: the character is what says the whitespace should go. Whereas
+[Remove Multiple Spaces](src/rules/remove-multiple-spaces.ts)
+collapses a run of spaces between any two non whitespace characters, and only the whitespace itself has to be
+unprotected, because anything at all can sit either side of it.
```
pjkaufman: Small typo:
```suggestion
unprotected, because anything at all can sit on either side of it.
```

src/rules/auto-correct-common-misspellings.ts
```diff
@@ -30,8 +32,24 @@ export default class AutoCorrectCommonMisspellings extends RuleBuilder<AutoCorre
...
+      if (range) {
+        const value = this.replaceWordWithCorrectCasing(match[0], options);
+        if (value !== match[0]) {
+          replacements.push({...range, value});
+        }
+      }
+    }
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This wording should probably be moved into `en.ts` and then use the `getLanguage` helpers to determine what to say here. That way it can be localized for each language.

pjkaufman: We may want to just go ahead and move this set of logic for checking for no overlap into a helper than can be reused across the rules.

src/rules/blockquote-style.ts
```diff
@@ -74,22 +98,15 @@ export default class BlockquoteStyle extends RuleBuilder<BlockquoteStyleOptions>
...
       [startOfLine, startOfIndex] = getStartOfLineWhitespaceOrBlockquoteLevel(newBlockquote, nextNewLine-1);
-      const startOfRestOfLine = startOfIndex + startOfLine.length+1;
-
-      let endOfRestOfLine = nextNewLine;
-      if (breakOutOfLoop) {
-        endOfRestOfLine++;
-      }
 
-      const restOfLine = newBlockquote.substring(startOfRestOfLine, endOfRestOfLine);
       // we need to ignore code and math blocks to prevent changing values in the display
```
pjkaufman: I think this comment should be removed since the associated code looks to have been removed:
```suggestion
```

src/utils/document-projection.ts
```diff
@@ -0,0 +1,121 @@
...
+  private readonly projectionEnds: number[] = [];
+
+  constructor(readonly source: string, replacements: ProjectionReplacement[]) {
+    const segments: string[] = [];
+    let sourceCursor = 0;
+    let projectionCursor = 0;
+    for (const {startIndex, endIndex, token} of replacements) {
+      if (!Number.isInteger(startIndex) || !Number.isInteger(endIndex) ||
+          startIndex < sourceCursor || endIndex <= startIndex || endIndex > source.length || token.length === 0) {
+        throw new RangeError('projection replacements must be sorted, nonempty and disjoint source ranges with nonempty tokens');
```
pjkaufman: This should probably be in en.ts and use the localization helpers to get the value so it can be represented in different languages.

src/utils/protected-ranges.ts
```diff
@@ -0,0 +1,480 @@
...
+   *
+   * Several rules mask a second set of types inside their own body, on top of the ones they
+   * declare. Going back through the context keeps that combination cached with all the others,
+   * rather than working the ranges out again every time the rule runs.
+   * @param {IgnoreType[]} ignoreTypes The further types to protect
+   * @return {ProtectedRanges} The regions of both sets of types
+   */
+  combinedWith(ignoreTypes: IgnoreType[]): ProtectedRanges {
+    if (!this.context || !this.ignoreTypes) {
+      throw new Error('protected ranges that did not come from a lint context cannot be combined with more ignore types');
```
pjkaufman: Another set of text that should be moved to the localization files and use the language helpers to get the final error message.

src/utils/protected-ranges.ts
```diff
@@ -0,0 +1,480 @@
...
+    return this.context.protectedRangesFor([...this.ignoreTypes, ...ignoreTypes]);
+  }
+
+  /**
+   * The cached decision view for these ignore types, owned by the same context as the ranges.
+   * @return {DocumentProjection} The projection with maps back to source offsets
+   */
+  projection(): DocumentProjection {
+    if (!this.context || !this.ignoreTypes) {
+      throw new Error('protected ranges that did not come from a lint context cannot provide a projection');
```
pjkaufman: Same comment here about the lcoalization.

src/rules/headings-start-line.ts
```diff
@@ -19,10 +21,19 @@ export default class HeadingStartLine extends RuleBuilder<HeadingStartLineOption
...
+    const replacements = collectUnprotectedRegexReplacements(
+        text, allHeadersRegex, protectedRanges, {
+          editRange: (match, startIndex) => ({startIndex, endIndex: startIndex + match[1].length, value: ''}),
+          // The heading marker must be visible, but its text can contain a placeholder.
+          guardRange: (match, startIndex) => ({startIndex, endIndex: startIndex + match[1].length + match[2].length + match[3].length}),
+        },
+    );
+    replacements.sort((a, b) => a.startIndex - b.startIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/heading-blank-lines.ts
```diff
@@ -22,23 +25,39 @@ export default class HeadingBlankLines extends RuleBuilder<HeadingBlankLinesOpti
...
+    for (const edit of getEditsBetween(projection.text, projectedText)) {
+      const range = projection.editRangeToSource(edit);
+      if (range) {
+        replacements.push({...range, value: edit.value});
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/header-increment.ts
```diff
@@ -71,8 +80,13 @@ export default class HeaderIncrement extends RuleBuilder<HeaderIncrementOptions>
 
       lastLevel = level;
 
-      return $1 + '#'.repeat(headingLevels[level - 1]) + $3 + $4 + $5;
-    });
+      replacements.push({...range, value: '#'.repeat(headingLevels[level - 1])});
+    }
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/footnote-after-punctuation.ts
```diff
@@ -18,10 +20,18 @@ export default class FootnoteAfterPunctuation extends RuleBuilder<FootnoteAfterP
...
     // Matches a footnote reference containing any text except newlines and the
     // terminating ].
-    return text.replace(/(\[\^[^\]]+\]) ?([,.;!:?])/gm, '$2$1');
+    const replacements = collectUnprotectedRegexReplacements(text, /(\[\^[^\]]+\]) ?([,.;!:?])/gm, protectedRanges, {
+      guardRange: (match, startIndex) => ({startIndex, endIndex: startIndex + match[0].length}),
+      editRange: (match, startIndex) => ({startIndex, endIndex: startIndex + match[0].length, value: match[2] + match[1]}),
+    });
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/file-name-heading.ts
```diff
@@ -41,25 +42,34 @@ export default class FileNameHeading extends RuleBuilder<FileNameHeadingOptions>
...
       header = '\n' + header;
     }
 
-    return insert(text, yaml_end, header);
+    const index = Math.min(yaml_end, projectedText.length);
+    const range = projection.editRangeToSource({startIndex: index, endIndex: index});
+    const replacements: textReplacement[] = range ? [{...range, value: header}] : [];
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/empty-line-around-tables.ts
```diff
@@ -19,8 +22,23 @@ export default class EmptyLineAroundTables extends RuleBuilder<EmptyLineAroundTa
...
+    for (const edit of getEditsBetween(projection.text, projectedText)) {
+      const range = projection.editRangeToSource(edit);
+      if (range) {
+        replacements.push({...range, value: edit.value});
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/consecutive-blank-lines.ts
```diff
@@ -20,8 +22,21 @@ export default class ConsecutiveBlankLines extends RuleBuilder<ConsecutiveBlankL
...
+    for (const match of projection.text.matchAll(multipleBlankLinesRegex)) {
+      const range = projection.editRangeToSource({startIndex: match.index, endIndex: match.index + match[0].length});
+      if (range) {
+        replacements.push({...range, value: '\n\n'});
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/capitalize-headings.ts
```diff
@@ -328,7 +330,45 @@ export default class CapitalizeHeadings extends RuleBuilder<CapitalizeHeadingsOp
...
+        const word = words[index];
+        addReplacement(cursor, word.index, index === 0 ? '' : ' ');
+        addReplacement(word.index, word.index + word[0].length, capitalizedWords[index]);
+        cursor = word.index + word[0].length;
+      }
+      addReplacement(cursor, match[0].length, '');
+    }
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/move-tags-to-yaml.ts
```diff
@@ -102,32 +104,50 @@ export default class MoveTagsToYaml extends RuleBuilder<MoveTagsToYamlOptions> {
...
+          removals.push({startIndex, endIndex, value: ''});
+        }
       }
-
-      return text;
-    });
+    }
+    removals.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (removals.some((replacement, index) => index > 0 && replacement.startIndex < removals[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/move-tags-to-yaml.ts
```diff
@@ -102,32 +104,50 @@ export default class MoveTagsToYaml extends RuleBuilder<MoveTagsToYamlOptions> {
...
+    const replacements: textReplacement[] = [];
+    for (const edit of getEditsBetween(projection.text, text)) {
+      const range = projection.editRangeToSource(edit);
+      if (range) {
+        replacements.push({...range, value: edit.value});
+      }
+    }
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/no-bare-urls.ts
```diff
@@ -25,29 +26,50 @@ export default class NoBareUrls extends RuleBuilder<NoBareUrlsOptions> {
...
+            replacements.push(replacement);
+          }
+        }
       }
     }
 
-    return text;
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/quote-style.ts
```diff
@@ -35,39 +36,37 @@ export default class QuoteStyle extends RuleBuilder<QuoteStyleOptions> {
...
 
-    return newText;
-  }
-  convertSmartSingleQuotesToStraightQuotes(text: string): string {
-    return text.replace(smartSingleQuoteRegex, '\'');
+    // The two quote families are disjoint, and converting either one does not change the
+    // letter/whitespace classification used by the other, so both read the original text.
+    replacements.sort((a, b) => a.startIndex - b.startIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/remove-empty-lines-between-list-markers-and-checklists.ts
```diff
@@ -19,26 +22,43 @@ export default class RemoveEmptyLinesBetweenListMarkersAndChecklists extends Rul
...
+    for (const edit of getEditsBetween(projection.text, projectedText)) {
+      const range = projection.editRangeToSource(edit);
+      if (range) {
+        replacements.push({...range, value: edit.value});
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/remove-empty-list-markers.ts
```diff
@@ -19,14 +22,32 @@ export default class RemoveEmptyListMarkers extends RuleBuilder<RemoveEmptyListM
...
+    for (const edit of getEditsBetween(projection.text, projectedText)) {
+      const range = projection.editRangeToSource(edit);
+      if (range) {
+        replacements.push({...range, value: edit.value});
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/remove-hyphenated-line-breaks.ts
```diff
@@ -18,8 +20,15 @@ export default class RemoveHyphenatedLineBreaks extends RuleBuilder<RemoveHyphen
...
-  apply(text: string, options: RemoveHyphenatedLineBreaksOptions): string {
-    return text.replace(/\b[-‐] \b/g, '');
+  apply(text: string, options: RemoveHyphenatedLineBreaksOptions, protectedRanges: ProtectedRanges): string {
+    const removeHyphen = (match: RegExpMatchArray, startIndex: number): textReplacement => ({
+      startIndex,
+      endIndex: startIndex + match[0].length,
+      value: '',
+    });
+    return replaceTextRanges(text, collectUnprotectedRegexReplacements(
+        text, /\b[-‐] \b/g, protectedRanges, {editRange: removeHyphen, guardRange: removeHyphen},
```
pjkaufman, line 30: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/remove-leftover-footnotes-from-quote-on-paste.ts
```diff
@@ -19,8 +21,18 @@ export default class RemoveLeftoverFootnotesFromQuoteOnPaste extends RuleBuilder
...
+  apply(text: string, options: RemoveLeftoverFootnotesFromQuoteOnPasteOptions, protectedRanges: ProtectedRanges): string {
+    // A placeholder's closing brace satisfies the non-digit anchor; only the suffix is removed.
+    const rangeForMatch = (match: RegExpMatchArray, startIndex: number) => ({startIndex: startIndex + match[1].length, endIndex: startIndex + match[0].length, value: ''});
+    const replacements = collectUnprotectedRegexReplacements(text, /(\D)[.,]\d+/g, protectedRanges, {
+      guardRange: rangeForMatch,
+      editRange: rangeForMatch,
+    });
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/remove-trailing-punctuation-in-heading.ts
```diff
@@ -21,22 +23,43 @@ export default class RemoveTrailingPunctuationInHeading extends RuleBuilder<Remo
...
         });
+        if (range) {
+          replacements.push({...range, value: ''});
+        }
+      }
+    }
+
+    replacements.sort((a, b) => a.startIndex - b.startIndex || a.endIndex - b.endIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/rules/space-after-list-markers.ts
```diff
@@ -18,14 +20,29 @@ export default class SpaceAfterListMarkers extends RuleBuilder<SpaceAfterListMar
...
+        text, /^(\s*\d+\.|\s*[-+*])[^\S\r\n]+/gm, protectedRanges, rangesForMatch,
     );
+    // Ordered markers were already handled above. Omitting that redundant alternative keeps
+    // the edits disjoint; normalizing marker whitespace cannot enable a new checkbox match.
+    replacements.push(...collectUnprotectedRegexReplacements(
+        text, /^(\s*[-+*]\s+\[[ xX]\])[^\S\r\n]+/gm, protectedRanges, rangesForMatch,
+    ));
+    replacements.sort((a, b) => a.startIndex - b.startIndex);
+    if (replacements.some((replacement, index) => index > 0 && replacement.startIndex < replacements[index - 1].endIndex)) {
+      throw new Error('Rule replacements must be ordered and non-overlapping');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

src/utils/mdast.ts
```diff
@@ -517,15 +655,37 @@ function addOrReplaceLineEnding(paragraphLine: string, indicator: LineBreakIndic
...
   }
 
+  const replacements: textReplacement[] = [];
+  const boundaryReplacements = new Map<string, textReplacement>();
+  const addBoundaryReplacement = (startIndex: number, endIndex: number, value: string): void => {
+    const key = `${startIndex}:${endIndex}`;
+    const existing = boundaryReplacements.get(key);
+    if (existing) {
+      if (existing.value != value) {
+        throw new Error('Paragraphs sharing a gap must agree on its replacement');
```
pjkaufman: This should be moved to the localization file to allow for better language support rather than just having errors in English.

pjkaufman (changes requested):
The vast majority of this seems fine to me. If I recall correctly the issues I noticed on the PR are as follows:
- There is English error text that should be moved to `en.ts` and retrieved via `getTextInLanguage`
- The checks for replacement overlaps seems common enough to warrant moving the logic to its own helper function that can be called.
- Some minor leftover comments and English grammar error

I can't say I follow all of the changes, but I think I can follow enough to be able pick things up once I work with things.

saberzero1:
Thanks — all three points are addressed in 438e1f0.

**The overlap check is now a helper.** `applyNonOverlappingReplacements` lives in `src/utils/text-edits.ts` next to the other edit utilities, and replaces the sort-assert-apply block that nineteen rules had each grown their own copy of. Twenty three call sites use it now. The assertion is kept rather than relaxed — it caught real bugs while this work was being done.

Doing that before the localization turned out to matter: most of the flagged strings were copies of the same overlap message, so collapsing them first left **five** distinct messages instead of twenty four.

**Those five are now in `en.ts`** under `logs`, read through `getTextInLanguage` like the rest of the project's text. Only `en.ts` gains entries, since the helpers fall back to English for a missing key — and there is a test that resolves every new key across all twenty four registered locales, so a mistyped key fails a test rather than only the type checker.

While moving them I noticed two began with a lowercase letter and one wrote "nonempty"; since these surface to users in a notice, I fixed those too.

**The leftover comment and the typo are gone.** The blockquote-style comment is deleted. "can sit either side" is now "can sit **on** either side" — fixed in `docs/docs/contributing/adding-a-rule.md` and `CONTRIBUTING.md` regenerated from it, since editing the generated file directly would have been overwritten.

One deliberate omission: `npm run docs` also rewrites `README.md` and `docs/docs/settings/{content,heading,yaml}-rules.md`, replacing real default values with `null`. **That drift already exists on master** — regenerating there moves `yaml-rules.md` by 520 lines. It comes from the `TextAreaOptionBuilder.buildOption` arity noted in the PR description, and committing it here would look like this PR degraded the docs. I have left those files untouched so the fix can go with the arity change instead. Happy to include it if you would rather have it in one go.

Gates: 82 suites / 1,780 tests pass, and a clean checkout with no fixture configured passes too. ESLint silent, `tsc` still at the 60 pre-existing errors, build passes, parse count still 15, and the differential corpus is byte-identical to pre-refactor output.

pjkaufman (approved):
Looks good to me. Thanks you for doing this work @saberzero1!

